PR #6, opened 2018-10-31
http: fix callers for when nested inside other methods in apps

adamdecaf:
Essentially, we want to be able to wrap InternalError inside other
functions in our apps, but the file-level information is nice to have.

Previously, the file information would be http.go:XX of the app, not
the route func we'd expect.


adamdecaf:
This adds a `source=http.go:103` to each log, but go-kit's log adds this (it's incorrect in this case).

Maybe I should patch this upstream?

PR #17, opened 2018-12-03
http: simplify CORS, fix content-type bug

adamdecaf:
Issue: https://github.com/moov-io/moov-io/issues/11

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -83,15 +83,14 @@ func AddCORSHandler(r *mux.Router) {
 			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
-		SetAccessControlAllowHeaders(w, r)
+		SetAccessControlAllowHeaders(w, r.Header.Get("Origin"))
 		w.WriteHeader(http.StatusOK)
 	})
 }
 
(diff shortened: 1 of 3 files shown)


http/server.go
```diff
@@ -101,11 +100,6 @@ func SetAccessControlAllowHeaders(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Access-Control-Allow-Methods", "GET,POST,PATCH,DELETE")
 		w.Header().Set("Access-Control-Allow-Headers", "Cookie,X-User-Id,X-Request-Id,Content-Type")
 		w.Header().Set("Access-Control-Allow-Credentials", "true")
-
-		// Don't overwrite content-type
-		if v := w.Header().Get("Content-Type"); v == "" {
-			w.Header().Set("Content-Type", "text/plain")
```
adamdecaf: We were setting this to try and stay within the preflight limits (i.e. not to require a preflight request), but that's already broken by sending credentials across domains.

tl;dr there's no point and this caused a bug in rendering ACH file contents

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

Issue #23, opened 2019-01-04
Move ParseError / ErrorList from moov-io / ach into base library for re-use 


bkmoovio:
There will be a need to re-use Parse Error / ErrorList for moov-io api's, packages, etc.

Copy ParseError / ErrorList from moov-io / ach into base library for re-use.

Put in root of github.com/moov-io/base and the it can be imported from other packages:  base.ErrorList

PR #25, opened 2019-01-07
Add parse error and error list

bkmoovio:
https://github.com/moov-io/base/issues/23

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,3 +1,7 @@
+// Copyright 2019 The Moov Authors
+// Use of this source code is governed by an Apache License
+// license that can be found in the LICENSE file.
+
 package base
 
 import (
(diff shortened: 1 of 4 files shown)


error.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (e ParseError) Error() string {
+	return fmt.Sprintf("%T %s", e.Err, e.Err)
+}
+
+// ErrorList represents an array of errors which is also an error itself.
+type ErrorList []error
+
+// Add appends err onto the ErrorList. Errors are kept in order.
```
adamdecaf:
```suggestion
// Add appends err onto the ErrorList. Errors are kept in append order.
```

PR #79, opened 2020-09-03
Added GetSkipAndCount() for reuse in pagination

jmbrown412:
Added a helper method GetSkipAndCount() to get skip and count params for resuse in pagination in Moov apis.

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -123,21 +123,21 @@ func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool, errors 
 	skipVal := r.URL.Query().Get("skip")
 	var err error
 	skip, err = strconv.Atoi(skipVal)
+	if err != nil && len(skipVal) > 0 {
+		errors = append(errors, err)
+		skip = 0
+	}
 	skip = int(math.Min(float64(skip), 10000))
 	skip = int(math.Max(0, float64(skip)))
-	if err != nil && skip == 0 && skipVal != "" {
-		errors = append(errors, err)
-		skip = -1
-	}
 
 	countVal := r.URL.Query().Get("count")
-	count, _ = strconv.Atoi(countVal)
+	count, err = strconv.Atoi(countVal)
+	if err != nil && len(countVal) > 0 {
+		errors = append(errors, err)
+		count = 0
+	}
 	count = int(math.Min(float64(count), 200))
 	count = int(math.Max(0, float64(count)))
-	if err != nil && count == 0 && countVal != "" {
-		errors = append(errors, err)
-		count = -1
-	}
 
 	exists = skipVal != "" || countVal != ""
 	return skip, count, exists, errors
--- a/http/server.go
+++ b/http/server.go
@@ -119,12 +119,14 @@ func GetUserID(r *http.Request) string {
 // skip is the number of records to pass over before starting a search
 // count is the number of records to retrieve in the search
 // exists indicates if skip or count was passed into the request URL
-func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool) {
+func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool, errors []error) {
 	skipVal := r.URL.Query().Get("skip")
-	skip, _ = strconv.Atoi(skipVal)
+	var err error
+	skip, err = strconv.Atoi(skipVal)
 	skip = int(math.Min(float64(skip), 10000))
 	skip = int(math.Max(0, float64(skip)))
-	if skip == 0 && skipVal != "" {
+	if err != nil && skip == 0 && skipVal != "" {
+		errors = append(errors, err)
 		skip = -1
 	}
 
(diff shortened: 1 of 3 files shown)


http/server.go
```diff
@@ -112,3 +113,17 @@ func GetRequestID(r *http.Request) string {
 func GetUserID(r *http.Request) string {
 	return strx.Or(r.Header.Get("X-User"), r.Header.Get("X-User-Id"))
 }
+
+// ReadSkipAndCount returns the skip and count pagination values from the query
```
adamdecaf, line 117:
```suggestion
// GetSkipAndCount returns the skip and count pagination values from the query
```

adamdecaf: Can we add a sentence about what "skip" and "count" mean to the UI or a calling client?

http/server.go
```diff
@@ -112,3 +113,17 @@ func GetRequestID(r *http.Request) string {
...
+func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool) {
+	skip, err := strconv.Atoi(r.URL.Query().Get("skip"))
+	if err != nil {
+		skip = 0
+	}
+	count, err = strconv.Atoi(r.URL.Query().Get("count"))
```
adamdecaf, line 123: Can we introduce a hard limit? 200 was what we agreed to in slack. `math.Min(count, 200)`

http/server.go
```diff
@@ -112,3 +113,17 @@ func GetRequestID(r *http.Request) string {
 func GetUserID(r *http.Request) string {
 	return strx.Or(r.Header.Get("X-User"), r.Header.Get("X-User-Id"))
 }
+
+// ReadSkipAndCount returns the skip and count pagination values from the query
+func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool) {
+	skip, err := strconv.Atoi(r.URL.Query().Get("skip"))
```
adamdecaf, line 119: Can we cap this? 10,000?

http/server.go
```diff
@@ -112,3 +113,17 @@ func GetRequestID(r *http.Request) string {
...
+	if err != nil {
+		skip = 0
+	}
+	count, err = strconv.Atoi(r.URL.Query().Get("count"))
+	if err != nil {
+		count = 0
+	}
+	exists = skip > 0 || count > 0
```
adamdecaf, line 127: I was thinking `exists` is only `true` if either skip or count are specified in the URL. We might want to know they're invalid values that got converted to 0.

http/server.go
```diff
@@ -112,3 +113,17 @@ func GetRequestID(r *http.Request) string {
...
+	return skip, count, exists
```
adamdecaf, line 128: Just realized we should prevent negative numbers. `math.Max(0, skip)` (same for count)

adamdecaf (commented):
Sorry Jordan for bombing this with a lot of comments.

http/server.go
jmbrown412: Done

http/server.go
jmbrown412: Done

http/server.go
jmbrown412: Done

http/server.go
jmbrown412: Done

http/server.go
jmbrown412: Done

http/server.go
```diff
@@ -112,3 +114,31 @@ func GetRequestID(r *http.Request) string {
...
+// count is the number of records to retrieve in the search
+// exists indicates if skip or count was passed into the request URL
+func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool, errors []error) {
+	skipVal := r.URL.Query().Get("skip")
+	var err error
+	skip, err = strconv.Atoi(skipVal)
+	skip = int(math.Min(float64(skip), 10000))
+	skip = int(math.Max(0, float64(skip)))
+	if err != nil && skip == 0 && skipVal != "" {
+		errors = append(errors, err)
```
adamdecaf, line 129: I think we should return the first error. If we can't parse skip/count then we should likely fail the request.

jmbrown412: I updated it to append the err only if they passed in a value so we won't return errors if they didn't pass in anything.

http/server.go
```diff
@@ -112,3 +114,42 @@ func GetRequestID(r *http.Request) string {
...
+func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool, errors []error) {
+	skipVal := r.URL.Query().Get("skip")
+	var err error
+	skip, err = strconv.Atoi(skipVal)
+	if err != nil && len(skipVal) > 0 {
+		errors = append(errors, err)
+		skip = 0
+	}
+	// TODO - Add this to a config somewhere?
+	// Cap skip at 10000
```
adamdecaf, line 131: I don't think we care about these being configurable. They're going to be hardcoded across services.

jmbrown412: Makes sense. Pushed.

http/server.go
adamdecaf: I don't understand why we're returning `[]error` though. Can we just return the first `error`?

jmbrown412: Yeah, that was confusing. I was trying to be too informative. Refactored to just return an err.

adamdecaf (approved):

PR #81, opened 2020-09-24
add stime pkg from moov-io/identity`

vxio:
I suggest we rename `stime` to `timetest` to indicate that this package should only be used in testing; similar to `net/httptest`. Let me know what you think.

Related issue https://github.com/moov-io/identity/issues/60

### DIFF
--- a/stime/static_time.go
+++ b/stime/static_time.go
@@ -1,0 +1,36 @@
+package stime
+
+import (
+	"time"
+)
+
+type StaticTimeService interface {
+	Change(update time.Time) time.Time
+	Add(d time.Duration) time.Time
+	TimeService
+}
+
+type staticTimeService struct {
+	time time.Time
+}
+
+func NewStaticTimeService() StaticTimeService {
+	return &staticTimeService{
+		time: time.Now().In(time.UTC).Round(time.Second),
+	}
+}
+
+// DeleteInvite - Delete an invite that was sent and invalidate the token.
+func (s *staticTimeService) Now() time.Time {
+	return s.time
+}
+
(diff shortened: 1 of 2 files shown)


adamdecaf:
I like `timetest`

InfernoJJ:
Unlike `net/httptest` this is used in production code. Except that its used not in testing too to get the current time of the system. I'm not onboard with that one. I called it `stime` for something like `service time` or `time service` variations on this I could get behind though.

timetest/system_time.go
```diff
@@ -0,0 +1,20 @@
...
+	Now() time.Time
+}
+
+type timeService struct{}
+
+func NewSystemTimeService() TimeService {
+	return &timeService{}
+}
+
+// DeleteInvite - Delete an invite that was sent and invalidate the token.
```
jmbrown412, line 17: Is this comment correct for the method? Looks like some left over copy pasta.

timetest/static_time.go
```diff
@@ -0,0 +1,36 @@
...
+	time time.Time
+}
+
+func NewStaticTimeService() StaticTimeService {
+	return &staticTimeService{
+		time: time.Now().In(time.UTC).Round(time.Second),
+	}
+}
+
+// DeleteInvite - Delete an invite that was sent and invalidate the token.
```
jmbrown412, line 23: Copy pasta?

jmbrown412 (approved):
Looks like there are some comments that are possibly left over copy pasta but LGTM otherwise.

vxio:
@InfernoJJ good point; I didn't realize `TimeService` was being used outside of `StaticTimeService`.

I'll change the package name back to `stime`.

PR #83, opened 2020-09-24
add database pkg from moov-io/identity

vxio:
Related issue: https://github.com/moov-io/identity/issues/60


adamdecaf:
Did this fail because there were no migrations to run?

vxio:
I'm not sure why that test failed; ran fine on my machine.

adamdecaf:
Restarted that test. Let's see what happens.

vxio:
@adamdecaf you were right, builds were failing because of no migrations folder in `base` repo; added a quick fix.

vxio:
Builds still failing with:
`dial tcp [::1]:4306: connect: connection refused--- FAIL: Test_NewAndMigration_MySql (0.00s)`

I think Docker Build isn't setup for this repo yet.

adamdecaf:
We probably need to setup docker image launching in this test. Identity diverted from Customers/PayGate in requiring `docker-compose up` rather than using `dockertest`. 

https://github.com/moov-io/customers/blob/master/internal/database/mysql.go#L222

database/mysql.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, err
+	}
+	db.SetMaxOpenConns(maxActiveMySQLConnections)
+
+	// Check out DB is up and working
+	if err := db.Ping(); err != nil {
+		return nil, err
+	}
+
+	// Migrate our database
```
alovak, line 70: remove this line?

database/migrator.go
```diff
@@ -0,0 +1,70 @@
...
+	log.Info().Log("Running Migrations")
+	_ = pkger.Include(config.migrationsDir)
+
+	driver, err := GetDriver(db, config)
+	if err != nil {
+		return err
+	}
+
+	m, err := migrate.NewWithDatabaseInstance(
+		"pkger:///migrations/",
```
alovak, line 32: This one should use `config.migrationsDir` (most probably `path.Base(config.migrationsDir)`)

From migrate package:

```go
pkger.Include("/module/path/to/migrations")
m, err := migrate.New("pkger:///module/path/to/migrations", ...)
```

database/database_test.go
```diff
@@ -0,0 +1,51 @@
...
+	config := DatabaseConfig{
+		DatabaseName: "identity",
+		MySql: &MySqlConfig{
+			Address:  "tcp(localhost:4306)",
+			User:     "identity",
+			Password: "identity",
+		},
+	}
+
+	_, close, err := NewAndMigrate(config, nil, nil)
```
alovak, line 33: I'm curious how does it work if we don't have mysql instance running and we do not expect error here?

database/mysql.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (r *TestMySQLDB) Close() error {
+	r.shutdown()
+
+	// Verify all connections are closed before closing DB
+	if conns := r.DB.Stats().OpenConnections; conns != 0 {
+		panic(fmt.Sprintf("found %d open MySQL connections", conns))
+	}
+
+	r.container.Close()
```
alovak, line 123: I would still close container even if we have opened connections.

database/sqlite.go
```diff
@@ -0,0 +1,153 @@
...
+
+	db, err := sql.Open("sqlite3", s.path)
+	if err != nil {
+		return nil, err
+	}
+	if err := db.Ping(); err != nil {
+		return db, err
+	}
+
+	// Migrate our database
```
alovak, line 59: ✂ comment

database/database_test.go
alovak: I missed your comment on this.

adamdecaf (approved):

adamdecaf:
Looks good, but we're still failing because MySQL isn't running in tests. 

https://github.com/moov-io/base/pull/83#issuecomment-698612913

PR #92, opened 2020-10-02
accept variable number of key, val in log.WithKeyValue

alovak:
at least I need this change to replace go-kit/log with base/log here: https://github.com/moov-io/base/blob/master/http/response.go#L61

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -13,7 +13,7 @@ import (
 type Logger interface {
 	With(ctxs ...Context) Logger
 	WithMap(mapCtx map[string]string) Logger
-	WithKeyValue(key, value string) Logger
+	WithKeyValue(keyvals ...string) Logger
 
 	Info() Logger
 	Error() Logger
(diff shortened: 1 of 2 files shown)


log/logger.go
```diff
@@ -94,10 +94,25 @@ func (l *logger) WithMap(mapCtx map[string]string) Logger {
...
+		keyvals = append(keyvals, log.ErrMissingValue.Error())
+	}
+
+	// split into key/value pars
+	for i, key := range keyvals {
+		// skip for values as they have odd index
+		if i%2 != 0 {
+			continue
+		}
+
```
vxio, line 110: To skip odd indexes, you could change this to 
```go
for (i := 0; i < len(keyvals); i += 2) {...}
```

log/logger.go
```diff
@@ -94,10 +94,25 @@ func (l *logger) WithMap(mapCtx map[string]string) Logger {
...
+		value := keyvals[i+1]
```
vxio, line 111: I suggest inlining `value`

vxio (commented):
Nice, I like this change.

vxio (approved):

PR #93, opened 2020-10-06
Revert "accept variable number of key, val in log.WithKeyValue"

alovak:
Reverts moov-io/base#92

Motivation for this revert is to reduce the chance of passing wrong amount of arguments to the method and getting broken logs. As example,

```go
log.WithKeyValue("one", "1", "two", "2", "three", "3")
```
Say you miss the "1" in that line and now your indexes are all off. You'd end up with "one=two" "2=three" in logs. 

Versus with fixed number of arguments (this PR) you are constrainted by the compiler to not mess it up:

```go
log.WithKeyValue("one", "1").WithKeyValue("two", "2")
```
Also, if you try to add 3 like this:

```go
log.WithKeyValue("three", ) 
```
it would compiler error.

Based on discussion with @InfernoJJ

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -13,7 +13,7 @@ import (
 type Logger interface {
 	With(ctxs ...Context) Logger
 	WithMap(mapCtx map[string]string) Logger
-	WithKeyValue(keyvals ...string) Logger
+	WithKeyValue(key, value string) Logger
 
 	Info() Logger
 	Error() Logger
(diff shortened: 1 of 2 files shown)


adamdecaf:
cc @InfernoJJ

log/logger.go
```diff
@@ -13,7 +13,7 @@ import (
 type Logger interface {
 	With(ctxs ...Context) Logger
 	WithMap(mapCtx map[string]string) Logger
-	WithKeyValue(keyvals ...string) Logger
+	WithKeyValue(key, value string) Logger
```
vxio, line 16: Nitpick:
Could we shorten this to `WithKV`?

alovak: I would make it `With` and `WithContext` depending on the frequency of usage :)

From my recent experience I used `WithKeyValue` a lot and zero times `With`.

alovak: @InfernoJJ what's your experience?

vxio: Agreed, looking at our `identity` service, `WithKeyValue()` is called a lot more frequently than `With(ctx ...Context)`. 
Renaming:
- `WithKeyValue` -> `With`
- `With` -> `WithContexts` 
would be great.

alovak: @InfernoJJ thoughts?

alovak: @vxio let's keep it as it. we can rename it pretty quickly if all projects together. Right now it may be a little bit complicated.

alovak:
@InfernoJJ please, approve :) It will take no time for you as I'm just reverting the code.

InfernoJJ (approved):

vxio (approved):
Dang, I liked this change.

alovak:
@vxio we will add:

```go
logger.With(log.Fields{
  "one": "1",
  "two": "2"
})
```

will it compensate your upset?

PR #94, opened 2020-10-06
base: add Logf method to logger

alovak:
Replacing logger in Customer (`go-kit/log` -> `base/log`) I face a lot of such use cases:

```go
logger.Log(fmt.Sprintf("updating address=%s for customer=%s", addressId, customerID))
```

`base/log` supports LogErrorF method, I think that it may be convenient and consistent to add Logf method to replace above with:
```go
logger.Log("updating address=%s for customer=%s", addressId, customerID)
```

I also think we should rename `LogErrorF` into `LogErrorf`. I can do this in separate PR. No one is using `base/log` now, so when Identity and other projects will switch to `base/log` developers will rename method. Thoughts?

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -20,6 +20,7 @@ type Logger interface {
 	Fatal() Logger
 
 	Log(msg string)
+	Logf(format string, a ...interface{})
 	LogError(msg string, err error) error
 	LogErrorF(format string, a ...interface{}) error
 }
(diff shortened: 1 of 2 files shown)


atonks2:
I think this is a great improvement. I also had the same thought regarding `LogErrorF` - that really should be `LogErrorf`.

atonks2 (approved):

InfernoJJ:
I'm against the lowercase `f` just because it doesn't fit naming conventions for shortened names and is just a hold out from older C.

alovak:
Closed in favor of https://github.com/moov-io/base/pull/99

PR #96, opened 2020-10-08
Database fixes

alovak:
This PR fixes 2 issues in base/database (which afaik is not used in other projects yet).

1. incorrect configuration of migration path
 
Expected behavior:
I can configure path to migrations files in `DatabaseConfig.migrationsDir` and they will be loaded from there during the migration

Actual behavior:
* There is no way to configure migrationsDir
* migrationsDir should be a module path (absolute path starting from the module root) and can't be treated as OS path. here is why:

Short: To embed migration files into binary we use [pkger](https://github.com/markbates/pkger/). `migrationsDir` can't be read by Pkger as Pkger uses static analysis and walks the AST of our application to find certain declarations such as pkger.Include and others. Because of this it can’t determine dynamic runtime values (which `migrationsDir` is).

Long (if you want to understand how pkger works): There are two modes of work in pkger.
**First,** if we tell pkger to read the file and it sees that this file was not embedded it tries to read it from disk. That's why when we run tests and pass migrations path as variable it works. Because /migration directory is there.
**Second,** if we build a binary (./bin/server) by providing migration file from configuration and we run our binary on the server where there is no directory /migrations it will try and fail to read it. In order to embed files into binary and be able to run it anywhere we should run `pkger` before we build our project. Like this (taken from [templater](https://github.com/moov-io/eng-project-templates/blob/master/.templates/go-service/Makefile#L84)):
```
$ pkger
$ go build -o ./bin/server cmd/server/*
```
`pkger` will walk the AST of our application and find all declarations with paths like `pkger.Include('/migrations')`. When it has all paths it will generate package.go file that will include the content of all scanned directories.

More about [how to use pkger](https://osinet.fr/go/en/articles/bundling-templates-with-pkger/).

2. no full clean up after dockertest

Expected behavior:
* when we run tests that launch MySQL instance in docker we expect all volumes to be removed after the test
* if test fails container should be stopped

Actual behavior:
* when test passes there is still volume with mysql instance container
* when test fails there is still running container

3. Use separate DB connection for service and migrations to avoid issue with not releasing 1 opened connection by migrator.

4. Unifies interface of `New` and `NewAndMigrate` functions

### DIFF
--- a/database/migrator.go
+++ b/database/migrator.go
@@ -55,19 +55,19 @@ func RunMigrations(logger log.Logger, config DatabaseConfig) error {
 }
 
 func GetDriver(db *sql.DB, config DatabaseConfig) (database.Driver, error) {
-	if config.MySql != nil {
-		return MySqlDriver(db)
+	if config.MySQL != nil {
+		return MySQLDriver(db)
 	} else if config.SQLite != nil {
-		return Sqlite3Driver(db)
+		return SQLite3Driver(db)
 	}
 
 	return nil, fmt.Errorf("database config not defined")
 }
 
-func MySqlDriver(db *sql.DB) (database.Driver, error) {
+func MySQLDriver(db *sql.DB) (database.Driver, error) {
 	return migmysql.WithInstance(db, &migmysql.Config{})
 }
 
-func Sqlite3Driver(db *sql.DB) (database.Driver, error) {
+func SQLite3Driver(db *sql.DB) (database.Driver, error) {
 	return migsqlite3.WithInstance(db, &migsqlite3.Config{})
 }
--- a/database/migrator.go
+++ b/database/migrator.go
@@ -50,7 +50,7 @@ func RunMigrations(logger log.Logger, db *sql.DB, config DatabaseConfig) error {
 func GetDriver(db *sql.DB, config DatabaseConfig) (database.Driver, error) {
 	if config.MySql != nil {
 		return MySqlDriver(db)
-	} else if config.SqlLite != nil {
+	} else if config.SQLite != nil {
 		return Sqlite3Driver(db)
 	}
 
--- a/database/migrator.go
+++ b/database/migrator.go
@@ -1,6 +1,7 @@
 package database
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 
@@ -14,7 +15,13 @@ import (
 	"github.com/moov-io/base/log"
 )
 
-func RunMigrations(logger log.Logger, db *sql.DB, config DatabaseConfig) error {
+func RunMigrations(logger log.Logger, config DatabaseConfig) error {
+	db, err := New(context.Background(), logger, config)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
 	logger.Info().Log("Running Migrations")
 
 	_ = pkger.Include("/migrations/")
@@ -37,12 +44,12 @@ func RunMigrations(logger log.Logger, db *sql.DB, config DatabaseConfig) error {
 	switch err {
 	case nil:
 	case migrate.ErrNoChange:
-		logger.Info().Logf("Database already at version")
+		logger.Info().Log("Database already at version")
 	default:
 		return logger.Fatal().LogErrorf("Error running migrations: %w", err).Err()
 	}
 
-	logger.Info().Logf("Migrations complete")
+	logger.Info().Log("Migrations complete")
 
 	return nil
 }
--- a/database/migrator.go
+++ b/database/migrator.go
@@ -3,7 +3,6 @@ package database
 import (
 	"database/sql"
 	"fmt"
-	"os"
 
 	"github.com/golang-migrate/migrate/v4"
 	"github.com/golang-migrate/migrate/v4/database"
@@ -16,12 +15,9 @@ import (
 )
 
 func RunMigrations(logger log.Logger, db *sql.DB, config DatabaseConfig) error {
-	if _, err := os.Stat(config.migrationsDir); os.IsNotExist(err) {
-		return fmt.Errorf("migrations directory=\"%s\" does not exist", config.migrationsDir)
-	}
+	logger.Info().Log("Running Migrations")
 
-	logger.Info().Logf("Running Migrations")
-	_ = pkger.Include(config.migrationsDir)
+	_ = pkger.Include("/migrations/")
 
 	driver, err := GetDriver(db, config)
 	if err != nil {
@@ -29,7 +25,7 @@ func RunMigrations(logger log.Logger, db *sql.DB, config DatabaseConfig) error {
 	}
 
 	m, err := migrate.NewWithDatabaseInstance(
-		fmt.Sprintf("pkger://%s", config.migrationsDir),
+		"pkger:///migrations/",
 		config.DatabaseName,
 		driver,
 	)
--- a/database/model_config.go
+++ b/database/model_config.go
@@ -1,10 +1,9 @@
 package database
 
 type DatabaseConfig struct {
-	MySql         *MySqlConfig
-	SqlLite       *SqlLiteConfig
-	DatabaseName  string
-	migrationsDir string
+	MySql        *MySqlConfig
+	SqlLite      *SqlLiteConfig
+	DatabaseName string
 }
 
 type MySqlConfig struct {
--- a/database/model_config.go
+++ b/database/model_config.go
@@ -2,7 +2,7 @@ package database
 
 type DatabaseConfig struct {
 	MySql        *MySqlConfig
-	SqlLite      *SqlLiteConfig
+	SQLite      *SQLiteConfig
 	DatabaseName string
 }
 
@@ -12,13 +12,13 @@ type MySqlConfig struct {
 	Password string
 }
 
-type SqlLiteConfig struct {
+type SQLiteConfig struct {
 	Path string
 }
 
 var InMemorySqliteConfig = DatabaseConfig{
 	DatabaseName: "sqlite",
-	SqlLite: &SqlLiteConfig{
+	SQLite: &SQLiteConfig{
 		Path: ":memory:",
 	},
 }
--- a/database/model_config.go
+++ b/database/model_config.go
@@ -1,12 +1,12 @@
 package database
 
 type DatabaseConfig struct {
-	MySql        *MySqlConfig
-	SQLite      *SQLiteConfig
+	MySQL        *MySQLConfig
+	SQLite       *SQLiteConfig
 	DatabaseName string
 }
 
-type MySqlConfig struct {
+type MySQLConfig struct {
 	Address  string
 	User     string
 	Password string
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -126,14 +126,12 @@ func (r *TestMySQLDB) Close() error {
 
 	err := r.DB.Close()
 	if err != nil {
-		fmt.Println("Failed to close MySQL database")
-		panic(err)
+		return err
 	}
 
 	err = r.container.Close()
 	if err != nil {
-		fmt.Println("Failed to close MySQL docker container")
-		panic(err)
+		return err
 	}
 
 	return nil
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -153,7 +153,7 @@ func CreateTestMySQLDB(t *testing.T) *TestMySQLDB {
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
 
-	db, err := NewAndMigrate(*config, logger, ctx)
+	db, err := NewAndMigrate(ctx, logger, *config)
 	if err != nil {
 		container.Close()
 		t.Fatal(err)
@@ -162,12 +162,7 @@ func CreateTestMySQLDB(t *testing.T) *TestMySQLDB {
 	// Don't allow idle connections so we can verify all are closed at the end of testing
 	db.SetMaxIdleConns(0)
 
-	shutdownFunc := func() {
-		cancelFunc()
-		db.Close()
-	}
-
-	return &TestMySQLDB{DB: db, container: container, shutdown: shutdownFunc}
+	return &TestMySQLDB{DB: db, container: container, shutdown: cancelFunc}
 }
 
 func RunMySQLDockerInstance(config *DatabaseConfig) (*DatabaseConfig, *dockertest.Resource, error) {
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -15,6 +15,7 @@ import (
 	kitprom "github.com/go-kit/kit/metrics/prometheus"
 	gomysql "github.com/go-sql-driver/mysql"
 	"github.com/ory/dockertest/v3"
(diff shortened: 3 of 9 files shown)


InfernoJJ (changes requested):
I don't understand the need for the `migrationsDir` configuration? Where would we use this?

I'm working to keep things easier and boring by saying we follow a specific format.

database/sqlite.go
```diff
@@ -129,7 +119,15 @@ func CreateTestSqliteDB(t *testing.T) *TestSQLiteDB {
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
 
-	db, err := sqliteConnection(log.NewNopLogger(), filepath.Join(dir, "paygate.db")).Connect(ctx)
+	// logger = log.NewNopLogger()
+	logger := log.NewDefaultLogger()
+
+	db, err := sqliteConnection(logger, filepath.Join(dir, "paygate.db")).Connect(ctx)
```
InfernoJJ, line 125: This is a copy and paste and it should probably be cleaned up. Notice the `paygate.db` sitting there. When I originally set this up it was copy and pasted from paygate and morphed to fit into `identity` and then coped into here. I think now that its in its final resting place we should clean this up. Also these `Test{MySQL|SQLite}DB structs aren't used in the templater code.

alovak:
@InfernoJJ 

> I don't understand the need for the `migrationsDir` configuration? Where would we use this?

it was somehow introduced and with this PR it's gone.

database/sqlite.go
alovak: During transition customers to "templater" I'll remove it. 

Btw, how does templater launche MySQL instance to test agains mysql db?

InfernoJJ: It loads it up via docker-compose. When you run `make test` or `make check` it runs `docker-compose up`

InfernoJJ:
@alovak ah lol I got my adds versus removes mixed up. Still weird that it got added because yeah it wouldn't work with `pkger`

database/sqlite.go
InfernoJJ: A single running external instance works great for this since we're building multi-tenancy it should maintain that one test using say tenant A shouldn't interfere with another test using say tenant B.

alovak: do we have an example of such mysql / testing / configuration?

right now in customers and paygate we use dockertest and it launches container.

InfernoJJ (approved):

InfernoJJ:
Yup the templates generates one for the repository tests. Just need to
incumbent the MySQL configuration in turn to turn it on.

adamdecaf (changes requested):
Does this require migrations exist in an app now?

alovak:
@adamdecaf yes. If you create `NewAndMigrate` or use `CreateTestMySQLDB` migrator will run migrations from /migrations directory.

database/mysql.go
```diff
@@ -113,12 +114,29 @@ type TestMySQLDB struct {
...
+	err := r.DB.Close()
+	if err != nil {
+		fmt.Println("Failed to close MySQL database")
+		panic(err)
+	}
+
+	err = r.container.Close()
+	if err != nil {
+		fmt.Println("Failed to close MySQL docker container")
+		panic(err)
```
vxio, line 136: Instead of calling `panic(err)` , we should `return err` and let the caller `panic` if they want to

adamdecaf:
@alovak what about when there are no migrations? Are callers supposed to check the amount of migrations prior?

InfernoJJ:
@adamdecaf then it won't be called and they can delete the code from the templater calling it.

alovak:
> what about when there are no migrations? Are callers supposed to check the amount of migrations prior?

@adamdecaf do you mean the case when there is a `/migrations` directory but there are no files in it?

database/migrator.go
```diff
@@ -16,20 +15,17 @@ import (
...
 
 func RunMigrations(log log.Logger, db *sql.DB, config DatabaseConfig) error {
-	if _, err := os.Stat(config.migrationsDir); os.IsNotExist(err) {
-		return fmt.Errorf("migrations directory=\"%s\" does not exist", config.migrationsDir)
-	}
-
 	log.Info().Log("Running Migrations")
-	_ = pkger.Include(config.migrationsDir)
+
+	_ = pkger.Include("/migrations/")
```
adamdecaf, line 20: Does this mean all projects need this directory? Do we need to remove it from projects to prevent double imports?

See: https://github.com/moov-io/access/blob/master/package.go#L10

database/model_config.go
```diff
@@ -1,10 +1,9 @@
 package database
 
 type DatabaseConfig struct {
-	MySql         *MySqlConfig
-	SqlLite       *SqlLiteConfig
-	DatabaseName  string
-	migrationsDir string
+	MySql        *MySqlConfig
+	SqlLite      *SqlLiteConfig
```
adamdecaf, line 5: This is called SQLite - https://sqlite.org/index.html 
I know it was misspelled before, but can we fix it?

database/mysql.go
```diff
@@ -113,12 +114,27 @@ type TestMySQLDB struct {
 func (r *TestMySQLDB) Close() error {
 	r.shutdown()
 
+	// ASK: not sure what do we check in the following.
+	// We don't always close sql.DB directly but call
+	// this Close method.
```
adamdecaf, line 119: I'd like to maintain this check. It's saved myself from a bug where i had an incorrect transaction setup in my repository which led to a connection left open (and eventually connection pool exhaustion).

adamdecaf: This check runs at the end of a `func TestXXX(t *testing.T)` method so the underlying connection(s) should already be closed.

migrations/001_create_tests.up.sql
```diff
@@ -0,0 +1 @@
+create table tests (id varchar(10))
```
adamdecaf, line 1: Will this be included in each project importing `github.com/moov-io/base/database`?

database/migrator.go
alovak: In reality this Include in runtime does nothing (returns the same accepted string).

This line works only if we run `pkger` binary inside this project. It will scan all files for "pkger.Include(...)" and pack found files into pkged.go.

When we import this package and in Customers (or any other project) and run `pkger` binary it will not scan this specific file (moov/base/database/migrator.go) and will not pack anything.

That's why we have to create kind of fake file like the file you mentioned (https://github.com/moov-io/access/blob/master/package.go#L10) to let pkger know what we want him to package.

In this PR I decided to make it different. I added following to makefile:

```
build:
   pkger -include /migrations
   go build ....
```

it seems instead of passing paths via -include option I need to follow our convention and create `package.go` file the same as in access project.

database/model_config.go
alovak: I'll change it!

database/mysql.go
alovak: The strange thing about this code is:
we call TestMySQLDB.Close() which checks if we closed DB and if yes then it will close DB :D 

```
# we want to close DB
func (r *TestMySQLDB) Close()
...
# did we closed DB?
if conns := r.DB.Stats().OpenConnections; conns != 0 {
  # no, we didn't close DB
  # panic
 ...
}

# close DB
err := r.DB.Close()
```

maybe I don't understand the flow or miss something here...

I may guess that some code, having access to TestMySQLDB.DB may close it before calling TestMySQLDB.Close(). But at least in one case I was not able to do this...

migrations/001_create_tests.up.sql
alovak: No. It just to make this package test pass (check that it really may migrate DB).

database/mysql.go
vxio: hmm we've been calling `db.Close()` but it doesn't look like it's necessary. The mention in code comments in `database/sql`:
```
// Close closes the database and prevents new queries from starting.
// Close then waits for all queries that have started processing on the server
// to finish.
//
// It is rare to Close a DB, as the DB handle is meant to be
// long-lived and shared between many goroutines.
func (db *DB) Close() error {
```

and I've read the connection pool handles reuse and closing.

vxio (approved):
Generally looks ok to me after you've addressed @adamdecaf 's comments

database/mysql.go
adamdecaf: We should `Close()` the Conn and DB objects in our tests. Yes, in production code it's rare to close an `*sql.DB`, but in tests it's common as cleanup.

alovak: Guys, the question here is not close or not to close.

But the code itself. It does not allow to call DB close if it's not closed 🤷‍♂️

I just want to remove the code that checks if everything was closed as I don't understand it and it seems counter-intuitive to me.

adamdecaf: This check helped me avoid bugs. It's a great check to include. I don't understand why we'd remove it.

adamdecaf: Here's the example of this check on `.OpenConnections` helped me diagnose a connection leak. Let's keep this code and prevent bugs in the future. 

https://github.com/moov-io/paygate/pull/344

database/migrator.go
adamdecaf: I'm worried about us requiring `pkger` being on a developers `PATH` in order to build PayGate. Sure, to create a docker image it makes sense, but even to run `go test` or `go run` type behaviors (via `make build`) we need `pkger` installed now?

database/mysql.go
alovak: From sql.go mentioned by @vxio :

// Close closes the **database** and prevents new queries from starting.
// Close then waits for all queries that have started processing on the server
// to finish.
//
// It is rare to Close a DB, as the DB handle is meant to be
// long-lived and shared between many goroutines.
func (db *DB) Close() error {
...
}

As I understand it:

* we connect to a database (we do this once in an app)
* DB driver maintains pool of connections (automatically, based on set max con, idle con, etc.)
* if there is no free connection for sql query it will automatically open a new connection (invisibly to user) and use it for the query and then keep it in the pool of connections
* we do not close connections from pool as we don't have control over them
* We close (global) connection to the database which will close all opened connections inside the pool after queries finish / or time out

I did some experiments to double check my understanding and show that our `Close` method should be changed as it does not work it supposed to work. It worked before because in most cased in our tests we close `sql.DB` before we call `TestMySQLDB.Close` method (and goal of `TestMySQLDB.Close` is exactly to close DB).

Here is first experiment:

```go
func TestMySQLExperiment1(t *testing.T) {
	db := CreateTestMySQLDB(t)
	db.Close()
}

# fails
# --- FAIL: TestMySQLExperiment1 (18.14s)
#panic: found 1 open MySQL connections [recovered]
#        panic: found 1 open MySQL connections
```

I would expect it to work.

second experiment to show that we do not control (open or close) connections in the pool:

```go
func TestMySQLExperiment2(t *testing.T) {
	db := CreateTestMySQLDB(t)
	defer db.Close()

	ticker := time.NewTicker(100 * time.Millisecond)
	done := make(chan bool)

	go func() {
		for {
			select {
			case <-done:
				return
			case <-ticker.C:
				fmt.Printf("(1) open connections: %d\n", db.DB.Stats().OpenConnections)
			}
		}
	}()

	performQuery := func() {
		for {
			rows, _ := db.DB.Query("select * from tests")
			rows.Close()
			time.Sleep(50 * time.Millisecond)
		}
	}

	go performQuery()
	go performQuery()
	go performQuery()

	time.Sleep(2 * time.Second)
	ticker.Stop()
	done <- true
}

# output
# go test -run TestMySQLExperiment2
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 4
# (1) open connections: 1
# (1) open connections: 4
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 4
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 1
# (1) open connections: 4
# (1) open connections: 1
# (1) open connections: 4
# --- FAIL: TestMySQLExperiment2 (20.20s)
# panic: found 4 open MySQL connections [recovered]
#         panic: found 4 open MySQL connections
```

From 2nd experiment we can see that we opened only one connection, but somehow got 4 opened and we also see that there is no way for us to close... 3 of them.

Finally,
* we need to remove commented lines from our `Close` method and
* just close DB together with docker container.

adamdecaf: That's interesting.. It seems odd to me we can't close the `*sql.DB` in tests, but that appears to be from the ordering of `defer` and `t.Cleanup` calls. I think `t.Cleanup` is called before before our `defer` calls if we were to get the call stack.

database/migrator.go
adamdecaf: We solved this by including the generated file in our OSS projects.

database/mysql.go
alovak: Experiment 3 (just create test mysql db and see how many open connections we have from the start -> we expect to have 0):

```go
func TestMySQLExperiment3(t *testing.T) {
	db := CreateTestMySQLDB(t)
        defer db.Close()
	ticker := time.NewTicker(100 * time.Millisecond)
	done := make(chan bool)
	go func() {
		for {
			select {
			case <-done:
				return
			case <-ticker.C:
				fmt.Printf("open connections: %d\n", db.DB.Stats().OpenConnections)
			}
		}
	}()
	time.Sleep(2 * time.Second)
	ticker.Stop()
	done <- true
}
```
the result:

image

alovak: Our assumption that we somehow have 0 open connections is wrong. I don't know why. Maybe it's because of db config we have right now (max open connections, max idle connections, etc.).

alovak: I found one place that increases opened connections from 0 to 1. It's `err = RunMigrations(logger, db, *config)` which opens connection. Now the question, why it does not close it.

Here is the discussion on this topic https://github.com/golang-migrate/migrate/issues/97
https://github.com/golang-migrate/migrate/issues/97#issuecomment-464445111

@adamdecaf the issue is with migrator that opens one more connection that should be closed (released). As it accepts "db" that created for the app we can't call "Close()" on it. Solution is to create to separate db instance for migrations and close it.

alovak: Solved by using [separate DB connection](https://github.com/moov-io/base/pull/96/files#diff-7c9b482ada7e6e60f6c5c4593eaed1bb967e0813d28b63e278a9649046636395R19) for migrator.

database/model_config.go
```diff
@@ -1,10 +1,9 @@
 package database
 
 type DatabaseConfig struct {
-	MySql         *MySqlConfig
-	SqlLite       *SqlLiteConfig
-	DatabaseName  string
-	migrationsDir string
+	MySql        *MySqlConfig
```
alovak, line 4: Should I rename it to MySQL? :)

adamdecaf: I would yes.

alovak: Done!

adamdecaf (approved):
Awesome work Pavel, nice!

PR #99, opened 2020-10-09
Log improvements proposal

InfernoJJ:


alovak:
@InfernoJJ

for me it's not obvious that calling `LogError` returns error. Maybe we can remove it and just keep `Log` which will return error if it was set before using `WithError`?

```
return nil, logger.WithError(e).Log("failed to create user")
```

alovak:
One more question, will `WithError(e)` set level to error?

alovak:
I really like the interface you offered! 👍

It would be great to get hear the feedback from other guys.

InfernoJJ:
I don't want to set the level to `Error` on `.WithError` because not every error is an error. Say they fail validation, this is an issue with the customer and if we wanted to log it out stating its an `Error` level is not correct because its no an issue with the service but with the customer and we'd log it as an `Info` or `Debug`

InfernoJJ:
I thought about removing `LogError` and have `Log` return err or nil if it not set, however this creates warnings at compile time about a return value not being used. I'm up for renaming the `LogError` method to something else like `Log1` for returning 1 value or something, but we do need both to insist on the intentions of what we're calling it for.

jmbrown412 (approved):

log/logger_impl.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	for _, c := range ctxs {
+		itemCtx := c.Context()
+		for k, v := range itemCtx {
+			combined[k] = v
+		}
+	}
+
+	return &logger{
```
vxio, line 69: We should reuse our instance of a `logger`.

Here we can do
```go
l.combined = combined
return l
```

log/logger_impl.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type logger struct {
+	writer  log.Logger
+	ctx     map[string]string
+	lastErr error
+}
+
+func (l *logger) Set(key, value string) Logger {
+	return l.With(Fields{
```
vxio, line 48: I don't think we need to call `With` here.

We can do
```go
l.ctx[key] = value
return l
```

log/logger_impl.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	return &logger{
+		writer:  l.writer,
+		ctx:     combined,
+		lastErr: l.lastErr,
+	}
+}
+
+func (l *logger) WithError(err error) Logger {
```
vxio, line 76: We shouldn't add this method since it forces the caller to set `"msg"` with a call to `Log()`.

For example, we should be able to do
```go
_, err := createCustomer()
if err != nil {
  err = logger.LogError(fmt.Errorf("creating customer: %v, err))
  // write err to resp body or whatever
}
```
which should output a log without `"msg"`
```json
{ 
"error": "...some error message..."
}
```

---- 
Using `WithError()` makes us do
```go
err = log.WithError(err).Log("message %w", err)
```
which duplicates information in the `"error"` and `"msg"` fields of the log.

log/logger_impl.go
```diff
@@ -0,0 +1,132 @@
...
+		keyvals[i] = k
+		keyvals[i+1] = v
+		i += 2
+	}
+
+	_ = l.writer.Log(keyvals...)
+}
+
+// LogError logs the error or creates a new one using the msg if `err` is nil and returns it.
+func (l *logger) LogError(format string, a ...interface{}) error {
```
vxio, line 123: I'm a bit confused about how this method is implemented. We shouldn't need to check if `l.lastErr` is nil or not.
We should set err in the `"error"` field and log that with anything else in `l.ctx`
```go
err := fmt.Errorf(format, a...)
l.Set("error", err)
l.writer.Log() // we use writer.Log() since "error" field is already set
return err
```

darwinz (approved):

log/logger_impl.go
adamdecaf: I'm a big fan of this

log/logger_impl.go
```diff
@@ -0,0 +1,132 @@
...
+	combined := make(map[string]string, len(l.ctx)+len(ctxs))
+
+	for k, v := range l.ctx {
+		combined[k] = v
+	}
+
+	for _, c := range ctxs {
+		itemCtx := c.Context()
+		for k, v := range itemCtx {
+			combined[k] = v
```
adamdecaf, line 65: This doesn't appear to support multiple concurrent calls, is that correct?

alovak:
> I thought about removing `LogError` and have `Log` return err or nil if it not set, however this creates warnings at compile time about a return value not being used. I'm up for renaming the `LogError` method to something else like `Log1` for returning 1 value or something, but we do need both to insist on the intentions of what we're calling it for.

@InfernoJJ I agree. Haven't thought about value not being used. Also I agree with @vxio that combination of both: WithError and LogError adds "error": "error message" together with "msg": "message".

log/logger_impl.go
InfernoJJ: You don't want to re-use it because the mutability would affect other instances of it and make it not thread safe anymore.

log/logger_impl.go
InfernoJJ: Mutating it without creating a new instance of it can cause issues with thread-safe-ness. Using `l.With` is a way of re-using the correct pattern for adding values.

log/logger_impl.go
InfernoJJ: The `.WithError` is such that we can return the exact error type along with logging it for `.LogError`. The intend is to what you mention is a rare case which I say should be used more. `msg` is to give context to the action being taken when it failed, while `err` is the failure itself.

For instance, I'm looking up an identity.
```
identity, err := repo.getIdentity(someID)
if err != nil {
  return nil, logger.WithError(err).Log("unable to find identity {someID}"))
}
```

The log output ion this case would be.
```
msg="unable to find identity 1324512345"
error="sql.NoRows"
```

msg is the context of the action being taken when ti failed.
error is the reasoning why it failed.

Having the error stored as the original type allows us to match it later on without string matching it. As not all errors are just a simple string type.

log/logger_impl.go
InfernoJJ: This is a naming convention issue I think. The point of the `.LogError` is to return the error that was set with `.WithError` anyting specified in the arguments for `.LogError` is the message of the log entry not the error.

log/logger_impl.go
vxio: Ah I see.

log/logger_impl.go
vxio: In your example, wouldn't you want to add that context to the error? Such that the error would be:
`unable to find identity with {id}: sql: no rows in result set`.

In most cases where we're logging (the http handler level) you're not going to get the original error back, it'll hopefully be wrapped with additional context as it travels up the call stack.

log/logger_impl.go
vxio: Could you give some examples of when we'd want to use this method?

log/logger_impl.go
InfernoJJ: At the `http handler level` you need the original error types intact for it to do the switches on. You don't want to change it. So in your example we'd be switching the `sql.NoRows` error into a standard `error` type. When it gets to the http layer and inspects the type in your example it will return a `500` http status code because its no longer the `sql.NoRows` type to return a `404`

vxio: That shouldn't be an issue using `errors.Is()`, `errors.As()`, or some other check.

Even on validations it'll be wrapped and output something like `invalid key: sql: dupliate key value`.

vxio (approved):

PR #101, opened 2020-10-12
database: WIP of using a shared MySQL instance with a database per-test

adamdecaf:
We've had issues with projects using MySQL for testing being slow to run all tests. Instead what if we stand up one MySQL instance and create a database for each test? That would be a lot faster than the ~20-30s spin-up time of MySQL. 

I started working on this two weeks ago, but haven't gotten it finished. Wanted to get thoughts from folks. 

The idea being that each test calls `TestMySQLConnection(t)` which returns an `*sql.DB` that has a database created under the name `t.Name()` with all migrations ran. Then after that test finishes a `t.Cleanup` hook automatically closes the connection. After all tests finish we teardown the MySQL container.

### DIFF
--- a/database/mysql_testdb.go
+++ b/database/mysql_testdb.go
@@ -1,0 +1,150 @@
+// Copyright 2020 The Moov Authors
+// Use of this source code is governed by an Apache License
+// license that can be found in the LICENSE file.
+
+package database
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/moov-io/base/docker"
+	"github.com/moov-io/base/log"
+	"github.com/ory/dockertest"
+)
+
+var (
+	mysqlInstance *MySQLTestDB = &MySQLTestDB{}
+)
+
+func init() {
+	// after the above mysqlInstance is initialized with a zero-value
+	// write-lock our mysqlInstance so tests trying to read (RLock()
+	// are prevented until the underlying container is started.
+	mysqlInstance.mu.Lock()
+
+	go func() {
+		// ugly hack to stall .Wait() and allow tests to startup
+		time.Sleep(1 * time.Second)
+
+		// Block until all tests are finished
+		mysqlInstance.waiters.Wait()
+
+		// Shutdown our root connection and container
+		mysqlInstance.conn.Close()
+		mysqlInstance.container.Close()
+
+		// // Verify all connections are closed before closing DB
+		// if conns := r.DB.Stats().OpenConnections; conns != 0 {
+		// 	panic(fmt.Sprintf("found %d open MySQL connections", conns))
+		// }
+	}()
+}
+
+type MySQLTestDB struct {
+	conn      *sql.DB
+	container *dockertest.Resource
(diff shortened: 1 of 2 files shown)


database/mysql_testdb.go
```diff
@@ -0,0 +1,150 @@
...
+		mysqlInstance.container.Close()
+
+		// // Verify all connections are closed before closing DB
+		// if conns := r.DB.Stats().OpenConnections; conns != 0 {
+		// 	panic(fmt.Sprintf("found %d open MySQL connections", conns))
+		// }
+	}()
+}
+
+type MySQLTestDB struct {
```
adamdecaf, line 48: The idea of this thing is to act as a singleton for all `*testing.T` instances and create a new database whenever requested. We're blocking with an RWMutex that starts off in a write-locked state (so all reads are blocked) until the MySQL container is initialized. Then we're allowed to read-lock and initialize a database from `t.Name()`.

adamdecaf: As tests request a connection `TestMySQLConnection(t)` they're added into the `sync.WaitGroup` as the background goroutine (that does final cleanup) waits for them to finish. Once all registered tests are finished we close out the database.

This would be cleaner if Go had `BeforeAll` and `AfterAll` hooks we could register into, but without that we're doing a bit of racing to step in the middle of executing tests.

jmbrown412:
This sounds like a great improvement.

alovak:
This will speedup our testing for sure! Great improvement! 💪

1. For https://github.com/moov-io/base/blob/8150342846d25a88c9631ce1ff9c912aaa4d8a21/database/database_test.go#L23 we need to know how to connect to DB as we test public method that accepts DatabaseConfig. If MySQLTestDB has something like `CurrentConfig` that returns information on how to connect to DB instance then it makes possible to run/pass this test.

2. When we change data inside database we need to clean it up after the test. How do you see it? Maybe you will find this useful https://github.com/DATA-DOG/go-txdb.

3. Adding [this option](https://github.com/moov-io/base/blob/443c66fab32ff698baee9b97342bb71f088985a8/database/mysql.go#L210) (I tested it, but if you double check it would be great) to dockertest will remove container volume when it's shut down.

database/mysql_testdb.go
vxio: A possible alternative: Use `TestMain(m *testing.M)` to run migrations then expose a method like `database.Cleanup(t *testing.T)` that will perform a teardown.

Having to call `database.Cleanup(t)` in the beginning of each test isn't too bad.

database/mysql_testdb.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+func init() {
+	// after the above mysqlInstance is initialized with a zero-value
+	// write-lock our mysqlInstance so tests trying to read (RLock()
+	// are prevented until the underlying container is started.
+	mysqlInstance.mu.Lock()
+
+	go func() {
+		// ugly hack to stall .Wait() and allow tests to startup
```
vxio, line 31: A blocking channel could work well here. Or a using another `sync.WaitGroup` to block and calling `wg.Done()` in the `once.Do()`

PR #102, opened 2020-10-13
refactor logging package and add helpers

vxio:
Added methods:
- `logger.Log(msg string)` which calls `logger.Logf(format string, args ...string)`
- `logger.LogError(err error) error` which calls `logger.LogErrorf(format string, args ...string)`

And moved interfaces and helper items to `logger.go` file to simplify package layout.

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -9,6 +9,13 @@ type Logger interface {
 	Error() Logger
 	Fatal() Logger
 
-	Logf(format string, a ...interface{})
-	LogErrorf(format string, a ...interface{}) error
+	Log(message string)
+	Logf(format string, args ...interface{})
+
+	LogError(error error) error
+	LogErrorf(format string, args ...interface{}) error
+}
+
+type Context interface {
+	Context() map[string]string
 }
--- a/log/logger_test.go
+++ b/log/logger_test.go
@@ -18,7 +18,7 @@ func Test_LogImplementations(t *testing.T) {
 func Test_Log(t *testing.T) {
 	a, buffer, log := Setup(t)
 
-	log.Logf("my message")
+	log.Log("my message")
 
 	a.Contains(buffer.String(), "my message")
 	a.Contains(buffer.String(), "level=info")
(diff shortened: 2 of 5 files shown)


log/logger.go
```diff
@@ -9,6 +15,59 @@ type Logger interface {
 	Error() Logger
 	Fatal() Logger
 
-	Logf(format string, a ...interface{})
-	LogErrorf(format string, a ...interface{}) error
+	Log(message string)
+	Logf(format string, args ...interface{})
```
vxio, line 19: We could use `Logf(format string, args ...interface{})` for cases where we'd want to pass in a string without args, but adding `Log(msg string)` avoids any subtle bugs where the caller meant to pass in an interpolated string but forgot to add in the args.

log/logger_test.go
```diff
@@ -124,7 +124,7 @@ func Test_Caller(t *testing.T) {
 	a, buffer, log := Setup(t)
 
 	log.Info().With(StackTrace).Logf("message")
-	a.Regexp(regexp.MustCompile(`caller_0=(.*?)(\/log\/logger_test\.go)`), buffer.String())
+	a.Regexp(regexp.MustCompile(`caller_\d=(.*?)(\/log\/logger_impl\.go)`), buffer.String())
```
InfernoJJ, line 127: We don't want to announce the logger_impl.go file. We want to filter this out.

InfernoJJ: This is where the filter is: https://github.com/moov-io/base/pull/102/files#diff-330686dc447b6523e8cbb02ae01e63ed53f3b214274ecc0da55c978965dda5ddR63

InfernoJJ (changes requested):

darwinz (approved):

alovak (approved):

log/logger_test.go
```diff
@@ -124,7 +124,7 @@ func Test_Caller(t *testing.T) {
 	a, buffer, log := Setup(t)
 
 	log.Info().With(StackTrace).Logf("message")
-	a.Regexp(regexp.MustCompile(`caller_0=(.*?)(\/log\/logger_test\.go)`), buffer.String())
+	a.Regexp(regexp.MustCompile(`caller_\d=(.*?)(\/log\/logger_test\.go)`), buffer.String())
```
InfernoJJ, line 127: We also want to make sure its `called_0` so the first element of the stack where log was called.

InfernoJJ (changes requested):

InfernoJJ (approved):

PR #105, opened 2020-10-15
Remove returned error from LogError method and refactor method implementations

vxio:
`LogError(err)` should not return an `error` since the error will not be changed.

Refactored logging implementations without affecting behavior.

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -12,7 +12,7 @@ type Logger interface {
 	Log(message string)
 	Logf(format string, args ...interface{})
 
-	LogError(error error) error
+	LogError(error error)
 	LogErrorf(format string, args ...interface{}) error
 }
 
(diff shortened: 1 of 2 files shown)


adamdecaf:
I thought this was done to allow

```
err = logger.LogError(fmt.Errorf("foo: %v", err))
```

darwinz (approved):

vxio:
> I thought this was done to allow
> 
> ```
> err = logger.LogError(fmt.Errorf("foo: %v", err))
> ```

@adamdecaf `log.Errorf()` still has that behavior. You don't need to return an error in `log.Error()` since you're not wrapping the error.

InfernoJJ (changes requested):
We should still return the `err` even if its not changed so we can log it, return it, and move on. Doesn't matter if we're wrapping it or not.

InfernoJJ:
If I didn't want to return the error I could just do `.Log(err.String()` `.LogError` exists because you can't overload it with `.Log` to get the returned error.

PR #112, opened 2020-10-20
Add timestamp field to all logs

### LINKED ISSUE #109: log: add timestamp key-value to all logs
We should add a timestamp to all logs using
```go
logger = log.With(logger, "ts", log.DefaultTimestampUTC)

logger.Log("msg", "hello")

// Output:
// ts=2016-01-01T12:34:56Z msg=hello
```

vxio:
All logs would have a key-value pair like `time=2020-10-20T05:09:49Z` in `UTC`

Fixes #109

### DIFF
--- a/log/logger_impl.go
+++ b/log/logger_impl.go
@@ -92,7 +92,7 @@ func (l *logger) Fatal() Logger {
 func (l *logger) Log(msg string) {
 	orig := []string{
 		"msg", msg,
-		"time", time.Now().UTC().Format(time.RFC3339),
+		"ts", time.Now().UTC().Format(time.RFC3339),
 	}
 
 	keyvals := make([]interface{}, (len(l.ctx)*2)+len(orig))
--- a/log/logger_impl.go
+++ b/log/logger_impl.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/go-kit/kit/log"
 )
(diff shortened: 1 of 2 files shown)


log/logger_impl.go
```diff
@@ -89,12 +90,17 @@ func (l *logger) Fatal() Logger {
 }
 
 func (l *logger) Log(msg string) {
-	keyvals := make([]interface{}, (len(l.ctx)*2)+2)
+	orig := []string{
+		"msg", msg,
+		"time", time.Now().UTC().Format(time.RFC3339),
```
vxio, line 95: I went with `"time"` here as the key. I like `ts` too but two extra characters avoid an acronym.

InfernoJJ: I'd leave it `ts` as our system already looks at it without additional configuration.

adamdecaf: I don't think Loki cares but we have seen ts already populated.

vxio: I'll change it to `ts`

log/logger_test.go
```diff
@@ -20,9 +21,19 @@ func Test_Log(t *testing.T) {
...
-
-	a.Contains(buffer.String(), "my message")
-	a.Contains(buffer.String(), "level=info")
+	got := buffer.String()
+
+	a.Contains(got, "my message")
+	a.Contains(got, "level=info")
+
+	// check valid timestamp
+	idx := strings.Index(got, "time=")
```
darwinz, line 30: Minor nitpicky thing, but you may want to set a variable for `"time="` since you're using in two places here

InfernoJJ (approved):

log/logger_test.go
vxio: Good point, I originally had it as a var but duplicated it for readability.. but I changed it back in my latest commit

PR #113, opened 2020-10-20
sqlite: reuse the db connection after running migrations

vxio:
Ensure that `InMemorySQLiteConfig` can be shared.

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -26,18 +26,13 @@ func NewAndMigrate(ctx context.Context, logger log.Logger, config DatabaseConfig
 	}
 
 	// run migrations first
-	db, err := RunMigrations(logger, config)
+	err := RunMigrations(logger, config)
 	if err != nil {
 		return nil, err
 	}
 
-	// In SQLite, we can reuse the same connection
-	if config.SQLite != nil {
-		return db, nil
-	}
-
 	// create DB connection for our service
-	db, err = New(ctx, logger, config)
+	db, err := New(ctx, logger, config)
 	if err != nil {
 		return nil, err
 	}
--- a/database/database.go
+++ b/database/database.go
@@ -31,12 +31,17 @@ func NewAndMigrate(ctx context.Context, logger log.Logger, config DatabaseConfig
 		return nil, err
 	}
 
-	// // create DB connection for our service
-	// db, err := New(ctx, logger, config)
-	// if err != nil {
-	// 	return nil, err
-	// }
-	//
+	// In SQLite, we can reuse the same connection
+	if config.SQLite != nil {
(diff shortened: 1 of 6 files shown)


adamdecaf:
Is this related to https://github.com/moov-io/base/issues/114 ?

database/database.go
```diff
@@ -25,17 +25,19 @@ func NewAndMigrate(ctx context.Context, logger log.Logger, config DatabaseConfig
...
-	}
-
 	// run migrations first
-	if err := RunMigrations(logger, config); err != nil {
+	db, err := RunMigrations(logger, config)
+	if err != nil {
 		return nil, err
 	}
 
+	// In SQLite, we can reuse the same connection
```
adamdecaf, line 34: Can we document this is because `:memory:` creates a different database on each init?

vxio:
@adamdecaf Yeah, this is kind of a 'hack' I wanted to test out due to the limitation of the `golang-migrate` library. Seems to work so far as all tests are passing.

alovak:
Yes. migrator with MySQL [opens new connection](https://github.com/golang-migrate/migrate/blob/master/database/mysql/mysql.go#L85) (and does not close it after), while SQLite does not (it runs queries with DB, not connection).

alovak:
If we need in-memory SQLite support then we can use this hack. Or we can pick an options from here: https://github.com/moov-io/base/issues/114

database/migrator.go
```diff
@@ -20,7 +20,6 @@ func RunMigrations(logger log.Logger, config DatabaseConfig) error {
 	if err != nil {
 		return err
 	}
-	defer db.Close()
```
vxio, line 23: I had to delete this line because calling `db.Close()` here will remove the InMemorySQLite db; the good news is no idle connections are left based on our tests.

database/model_config.go
```diff
@@ -19,6 +19,6 @@ type SQLiteConfig struct {
 var InMemorySqliteConfig = DatabaseConfig{
 	DatabaseName: "sqlite",
 	SQLite: &SQLiteConfig{
-		Path: ":memory:",
+		Path: "file::memory:?cache=shared",
```
adamdecaf, line 22: `file::` looks wrong - is that correct?

vxio: We need the `file::` to make it a public cache. Ref: https://www.sqlite.org/inmemorydb.html

database/migrator.go
InfernoJJ: You can still do a `db.Close()` on a Sqlite3 DB as it won't remove the DB until theirs still connections to it.

PR #115, opened 2020-10-20
remove global var InMemorySQLite and minor refactor

vxio:
- Removed global var `InMemorySqliteConfig` to avoid incorrect usage in tests; all services should call `CreateTestSQLiteDB()` for testing with sqlite
- Renamed instances of `Sqlite` to `SQLite`
- Improved tests to check proper behavior

### DIFF
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -23,6 +23,7 @@ func Test_NewAndMigration_SQLite(t *testing.T) {
 	}}
 
 	db, err := NewAndMigrate(context.Background(), nil, *config)
+	require.NoError(t, err)
 	defer db.Close()
 
 	require.NoError(t, err)
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -3,17 +3,29 @@ package database
 import (
 	"context"
 	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
 	"github.com/moov-io/base/docker"
 )
 
-func Test_NewAndMigration_SQLite3(t *testing.T) {
-	db, err := NewAndMigrate(context.Background(), nil, InMemorySqliteConfig)
-	if err != nil {
-		t.FailNow()
-	}
-	db.Close()
+func Test_NewAndMigration_SQLite(t *testing.T) {
+	dir, err := ioutil.TempDir("", "sqlite-test")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	config := &DatabaseConfig{SQLite: &SQLiteConfig{
+		Path: filepath.Join(dir, "tests.db"),
+	}}
+
+	db, err := NewAndMigrate(context.Background(), nil, *config)
+	require.NoError(t, err)
+	_, err = db.Query("select * from tests")
+	require.NoError(t, err)
 }
 
 func Test_NewAndMigration_MySql(t *testing.T) {
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -23,6 +23,8 @@ func Test_NewAndMigration_SQLite(t *testing.T) {
 	}}
 
 	db, err := NewAndMigrate(context.Background(), nil, *config)
+	defer db.Close()
+
 	require.NoError(t, err)
 	_, err = db.Query("select * from tests")
 	require.NoError(t, err)
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -10,6 +10,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/moov-io/base/docker"
 
 	kitprom "github.com/go-kit/kit/metrics/prometheus"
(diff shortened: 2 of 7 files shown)


database/sqlite.go
```diff
@@ -87,31 +87,29 @@ func sqliteConnection(logger log.Logger, path string) *sqlite {
 // TestSQLiteDB is a wrapper around sql.DB for SQLite connections designed for tests to provide
 // a clean database for each testcase.  Callers should cleanup with Close() when finished.
 type TestSQLiteDB struct {
-	DB *sql.DB
-
-	dir string // temp dir created for sqlite files
-
+	*sql.DB
```
vxio, line 90: Embedded this to avoid having to use `db.DB` when we want to pass on the db connection. In almost all cases where we're using this wrapper struct, we pass on the underlying `sql.DB` value.

```go
	db := database.CreateTestSqliteDB(t)
	repo := NewRepo(log.NewNopLogger(), db.DB)
         // turns into
	db := database.CreateTestSqliteDB(t)
	repo := NewRepo(log.NewNopLogger(), db)
```

database/sqlite.go
```diff
@@ -87,31 +87,29 @@ func sqliteConnection(logger log.Logger, path string) *sqlite {
...
 	shutdown func() // context shutdown func
 }
 
 func (r *TestSQLiteDB) Close() error {
 	r.shutdown()
 
 	// Verify all connections are closed before closing DB
-	if conns := r.DB.Stats().OpenConnections; conns != 0 {
-		panic(fmt.Sprintf("found %d open sqlite connections", conns))
+	if conns := r.DB.Stats().Idle; conns != 0 {
```
alovak, line 99: Why do we check for idle instead of open connections?

vxio: Hmm maybe doing `if conns := r.DB.Stats().OpenConnections; conns != 1` would make more sense; Because I'll always have 1 connection open before I close it in `L102` with `r.DB.Close()`

alovak: with `db.SetMaxIdleConns(0)` our DB will close all unused connections. If we still have opened connections it may mean that some query is still keeping it open/active and we missed `Close()` on statement or transaction.

When I faced this issue (with having 1 opened connection) it was because of that issue with migrator. After I used two DB instances I got 0 open connections...

database/database_test.go
```diff
@@ -3,17 +3,29 @@ package database
...
+func Test_NewAndMigration_SQLite(t *testing.T) {
+	dir, err := ioutil.TempDir("", "sqlite-test")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	config := &DatabaseConfig{SQLite: &SQLiteConfig{
+		Path: filepath.Join(dir, "tests.db"),
+	}}
+
+	db, err := NewAndMigrate(context.Background(), nil, *config)
```
alovak, line 25: db.Close is missed?

database/sqlite.go
vxio: Ah I see, so we need to call close on all transactions or queries, I'll fix this, thanks for explaining :)

alovak: Btw, Adam was able to find and fix leaking connections thanks to this open connections check. Initially I insisted we have to remove it %-)

adamdecaf: Yep. Having a non-zero amount of connections is a bug. Since this has come up twice now we should document the code to mention this (and perhaps the error message as well). 

See https://github.com/moov-io/paygate/pull/344 for more information

adamdecaf (approved):

database/sqlite.go
vxio: I'll update this PR and include some comments documenting the behavior

database/mysql.go
```diff
@@ -104,7 +104,7 @@ func mysqlConnection(logger log.Logger, user, pass string, address string, datab
 // TestMySQLDB is a wrapper around sql.DB for MySQL connections designed for tests to provide
 // a clean database for each testcase.  Callers should cleanup with Close() when finished.
 type TestMySQLDB struct {
-	DB *sql.DB
+	*sql.DB
```
vxio, line 109: Also embedded the db here.

database/sqlite.go
vxio: Follow-up, take a look at the changes in this commit: https://github.com/moov-io/base/pull/115/commits/c68e38a9ccfd858b717a684488bda417faf66176

database/mysql.go
```diff
@@ -104,19 +106,25 @@ func mysqlConnection(logger log.Logger, user, pass string, address string, datab
...
 func (r *TestMySQLDB) Close() error {
-	// Verify all connections are closed before closing DB
-	conns := r.DB.Stats().OpenConnections
-
 	r.shutdown()
 
+	// Verify all connections are closed before closing DB
+	if conns := r.DB.Stats().OpenConnections; conns != 0 {
+		require.FailNow(r.t, ErrOpenConnections{
+			Database:       "sqlite",
```
alovak, line 123:
```suggestion
			Database:       "mysql",
```

alovak (approved):
great improvement! I've suggested one tiny change sqlite -> mysql but expect this everything is good!

PR #120, opened 2020-10-30
database: optimizing mysql

vxio:
Draft/POC for @alovak and @adamdecaf from discussions at #119

Calling `db := CreateTestMySQLDB` does the following:
1. Get a mysql instance with the container name `mysql-test-db`
   - if the container doesn't exist, it will create one named `mysql-test-db`
   - we can add a config option if we want to use something other than `mysql-test-db`
1. Open a db connection and check that it works
1. Run migrations only once since it's wrapped in a `once.Do()`
1. Call `Teardown()` to ensure we're working with a clean DB

`dockertest` doesn't allow us to set the port number and uses a random one everytime, so we're using the `containerName` as the identifier.

Right now, parallel tests are not supported, so you shouldn't put `t.Parallel()` in tests where you're using `mysql`. Down the line, we can add support for booting up multiple mysql databases in the same instance as suggest by Adam in https://github.com/moov-io/base/pull/119#discussion_r514592858

 Here's a quick benchmark where we're calling `CreateTestMySQLDB()` in different tests. It'll run the teardown method each time in the same MySQL container.
image

Also, I adjusted the database.Config APIs a bit and certain things still need to be refactored/validated. And I need to test this against some db tables in `customers`

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -11,11 +11,11 @@ import (
 // New establishes a database connection according to the type and environmental
 // variables for that specific database.
 func New(ctx context.Context, logger log.Logger, config Config) (*sql.DB, error) {
-	switch config.Type {
-	case TypeMySQL:
-		return mysqlConnection(logger, config.MySQL).Connect(ctx)
-	case TypeSQLite:
-		return sqliteConnection(logger, config.SQLite.Path).Connect(ctx)
+	switch c := config.(type) {
+	case MySQLConfig:
+		return mysqlConnection(logger, c).Connect(ctx)
+	case SQLiteConfig:
+		return sqliteConnection(logger, c.Path).Connect(ctx)
 	}
 
 	return nil, fmt.Errorf("database config not defined")
--- a/database/database.go
+++ b/database/database.go
@@ -10,17 +10,18 @@ import (
 
 // New establishes a database connection according to the type and environmental
 // variables for that specific database.
-func New(ctx context.Context, logger log.Logger, config DatabaseConfig) (*sql.DB, error) {
-	if config.MySQL != nil {
-		return mysqlConnection(logger, config.MySQL.User, config.MySQL.Password, config.MySQL.Address, config.DatabaseName).Connect(ctx)
-	} else if config.SQLite != nil {
+func New(ctx context.Context, logger log.Logger, config Config) (*sql.DB, error) {
+	switch config.Type {
+	case TypeMySQL:
+		return mysqlConnection(logger, config.MySQL).Connect(ctx)
+	case TypeSQLite:
 		return sqliteConnection(logger, config.SQLite.Path).Connect(ctx)
 	}
 
 	return nil, fmt.Errorf("database config not defined")
 }
 
-func NewAndMigrate(ctx context.Context, logger log.Logger, config DatabaseConfig) (*sql.DB, error) {
+func NewAndMigrate(ctx context.Context, logger log.Logger, config Config) (*sql.DB, error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -9,8 +9,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-
-	"github.com/moov-io/base/docker"
 )
 
 func Test_NewAndMigration_SQLite(t *testing.T) {
@@ -18,9 +16,12 @@ func Test_NewAndMigration_SQLite(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(dir)
 
-	config := &DatabaseConfig{SQLite: &SQLiteConfig{
-		Path: filepath.Join(dir, "tests.db"),
-	}}
+	config := &Config{
+		Type: TypeSQLite,
+		SQLite: SQLiteConfig{
+			Path: filepath.Join(dir, "tests.db"),
+		},
+	}
 
 	db, err := NewAndMigrate(context.Background(), nil, *config)
 	require.NoError(t, err)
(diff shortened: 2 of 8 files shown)


alovak:
@vxio thanks for this PR! As I told you before I took at least one of your ideas from here: check if container is launched and use it instead of launching a new one.

For (new) Config I think that we need to use mysql.Config -> DSN. Without creating own configs and functions to format DSN...  but we can work on it somehow later.

For Database clean up. If it's one DB then we will not be able to run tests in parallel. That's why I think it's better to create new DB ('create database %s') and run migrations instead of cleaning single DB.

If we are not going to use parallel tests, then it may be even better to use DB transactions with https://github.com/DATA-DOG/go-txdb. I used [it already](https://github.com/moov-io/customers/pull/171) and it works absolutely great (9seconds vs 2m:30s). Right now with our improvements it's 30 seconds which is far better than 2:30 but not as great as 9 secods %-)

I appreciate your help with DB improvements! Thanks!

vxio:
@alovak Hey Pavel! quick comment from my phone but will try to be more in depth later.

We should focus on two things: speed and ensuring each test has a clean DB to work with (proper teardowns). I think we should choose sequential, parallel, or transactional tests based on that. We should benchmark all apporaches which should be relatively simple to do.

If parallel tests are noticably slower than sequential because of db setup, then we shouldnt go that route.


-------
Idea: One other apporach we can try down the line to make tests even faster is create a resource pool. For example create 8 mysql databases with 8 tokens, and tests can queue up for a token to use a db; you can do this with Go channels to enable parallel tests; when a test is done, it'll return the token to the resource pool and the db should run the proper teardown.

alovak:
Before we focus on things you mentioned we need to discuss one limitation that @adamdecaf mentioned: we should not depend on external DB. Maybe we can review this limitation and accept that external DB can be used together with "dockertest". I think we may require users to have MySQL installed and configured. I don't see why we can't do this :D

Also, there are different time costs:
1. launch MySQL server -> 14seconds
2. create DB and run migrations for it -> 2seconds

> @alovak Hey Pavel! quick comment from my phone but will try to be more in depth later.

> We should focus on two things: speed and ensuring each test has a clean DB to work with (proper teardowns). I think we should choose sequential, parallel, or transactional tests based on that. We should benchmark all apporaches which should be relatively simple to do.

Sequential or parallel tests. It's interesting, but our tests run in parallel even if we think they are sequential :D

When we run `go test ./...` it runs in **package list mode**. In this mode, `go test` compiles package files and  test files into **test binary** and does this for each packages listed on the command line. Then it runs **test binaries** in parallel based on the number of CPUs available. We can make `go test ./...` completely sequential by specifying the number of test binaries that can be run in parallel with _-p_ option: `go test -p 1 ./...` (source: `go help build`).
 
It means, that idea of sharing one global variable (MySQL container) for all tests does not work. It will be shared only within a single test binary.

> If parallel tests are noticably slower than sequential because of db setup, then we shouldnt go that route. We should limit the amount of db setup (creating container or running migrations) as much as we can since that tends to be the most expensive.

> Idea: One other apporach we can try down the line to make tests even faster is create a resource pool. For example create 8 mysql databases with 8 tokens, and tests can queue up for a token to use a db; you can do this with Go channels to enable parallel tests; when a test is done, it'll return the token to the resource pool and the db should run the proper teardown.

Interesting idea. But, I don't see how it will help us. Time to create resources in the pool will be the same as it takes right now. Also, DB preparation/clean up will still be an issue. Also, the complexity of managing the pool... In general I believe we will not get any significant speed improvement from resource pool.

Btw, with shared mysql instance + using single container we were able to reduce test time from 2m:30s to 30s.

I'm biased a little bit to the solution with transactions as I used it many times before, but this solution is the fastest as well (10 seconds). This solution has drawback as well. It requires DB management like (db create, db migrate) so we can prepare DB before we run tests. Partially it's done because we use https://github.com/golang-migrate/migrate. We can run migrations (prepare DB for testing) without any changes to our binaries. We just have to run: `migrate -source file://path/to/migrations -database mysql://moov:secret@tcp(localhost:3306)/test up`.

We may discuss it on monday @adamdecaf @vxio

adamdecaf (commented):
Cool, so we wipe all the tables and migrate for each test? Running tests sequentially allows this. Pretty cool idea!

vxio:
@alovak Running each test in a transaction could work since concurrency and "teardown"'s are built in. I've never tried it, but I remember bookmarking https://github.com/DATA-DOG/go-txdb!

> Then it runs test binaries in parallel based on the number of CPUs available.

Good point; if this is an issue, we could add `sync.Mutex` to protect the db.

@adamdecaf Yeah, migrations are only done once and all rows are cleared at the beginning of each test. The ability to run tests in parallel could be added in the future; but given the # of tests we have so far, performance gains might be negligible.

alovak:
@adamdecaf @vxio guys, our tests binaries run in parallel while tests inside binary run sequentially (untless t.Parallel()). It means that:
1. we can't use single DB with migrating/wiping out tables. test binaries that run in parallel will corrupt this DB.
2. we can use single DB for **test binaries** if tests inside binary will use transaction, but DB should be prepared (migrated) before we run tests (0s/migration as it's migrated before) <- super fast
3. we can use unique DB per **test binary** with running migrations and use transaction (+2s/migration*N of test binaries) <- fast
4. we can use unique DB per **test** (inside binary) with running migrations for each unique DB (+2s/migration*N of total DB tests) <- slow

All this will work with single DB server. It can be run from dockertest or externally on your machine (in whatever way you run it).

P.S. super fast, fast, slow - are relative and subjective measures :)

vxio:
> 1. we can't use single DB with migrating/wiping out tables. test binaries that run in parallel will corrupt this DB.

@alovak I'm not against trying out transactions, but the latest commit on this PR should protect against parallel tests interfering with each other

https://github.com/moov-io/base/pull/120/commits/49822b036f4259deb1a50f06b25d87798f80de82

alovak:
@vxio it will not. as test binaries will not share this mutex. it will be shared within one test binary.

vxio:
> @vxio it will not. as test binaries will not share this mutex. it will be shared within one test binary.

Hmm after doing a bit of testing, I think the mutex should be shares across all test binaries.

adamdecaf:
Can we close this?

vxio:
@adamdecaf yes definitely, thanks for heads up

PR #121, opened 2020-11-04
config changes to remove global vars

vxio:
NOT an actual PR. Used for code suggestions for #119

### DIFF
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


database/mysql.go
```diff
@@ -153,26 +149,51 @@ func CreateTestMySQLDB(t *testing.T) *TestMySQLDB {
...
+
+	dbURL := fmt.Sprintf("%s:%s@%s/%s",
+		mysqlConfig.User,
+		mysqlConfig.Password,
+		mysqlConfig.Address,
+		dbName,
+	)
+
+	var db *sql.DB
+	err = pool.Retry(func() error {
```
alovak, line 176: Vince, good idea. From one point of view this code makes things simple. And I like it. We rely on docker to handle concurrent calls to Docker API (create container and find container).

But somehow I feel that this code is hiding "concurrency aspect" and may create some false assumptions... for example `err = pool.Retry(func() error {` will be called for each test and at some point when container is ready some tests may wait extra time (5+ seconds) in `Retry` method (they are async and `Retry` has exponential backoff). With `sync.Once` we know that container is ready.

I'm not sure we are adding more value with further changes to base/database. I'll wait for comments from Adam on this and then decide.

vxio: Good point, I've updated the code to only call `pool.Retry` when the container is first created.

In general, we should try to avoid global variables when we can since they can introduce unexpected behavior and hide dependencies.

PR #126, opened 2020-11-16
Remove the interface{} from being a valid argument when deep down go-kit can error out.

InfernoJJ:
Remove the interface{} from being a valid argument when deep down go-kit can error out.

### DIFF
--- a/log/model_formats.go
+++ b/log/model_formats.go
@@ -1,0 +1,57 @@
+package log
+
+import (
+	"encoding/base64"
+	"time"
+)
+
+// Renderer is an interface to deal with typing problems of just having an interface{} as the acceptable parameters
+// Go-kit logging has a failure case if you attempt to throw any values into it.
+// This is a way to guard our developers from having to worry about error cases of the lower logging framework.
+type Renderer interface {
+	getValue() interface{}
+}
+
+type any struct {
+	value interface{}
+}
+
+func (a *any) getValue() interface{} {
+	return a.value
+}
+
+func String(s string) Renderer {
+	return &any{s}
+}
+
+func Int(i int) Renderer {
+	return &any{i}
+}
+
+func Float64(f float64) Renderer {
+	return &any{f}
+}
+
+func Bool(b bool) Renderer {
+	return &any{b}
+}
+
+func Duration(d time.Duration) Renderer {
+	return &any{d.String()}
+}
+
+func Timestamp(t time.Time) Renderer {
+	return TimestampFormat(t, time.RFC3339Nano)
+}
+
+func TimestampFormat(t time.Time, format string) Renderer {
+	return String(t.Format(format))
+}
+
+func ByteString(b []byte) Renderer {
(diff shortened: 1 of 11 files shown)


log/model_formats.go
```diff
@@ -0,0 +1,57 @@
...
+
+func Timestamp(t time.Time) Renderer {
+	return TimestampFormat(t, time.RFC3339Nano)
+}
+
+func TimestampFormat(t time.Time, format string) Renderer {
+	return String(t.Format(format))
+}
+
+func ByteString(b []byte) Renderer {
```
vxio, line 51:
```suggestion
func Bytes(b []byte) Renderer {
```

log/model_formats.go
```diff
@@ -0,0 +1,57 @@
...
+
+func Float64(f float64) Renderer {
+	return &any{f}
+}
+
+func Bool(b bool) Renderer {
+	return &any{b}
+}
+
+func Duration(d time.Duration) Renderer {
```
vxio, line 39:
```suggestion
func TimeDuration(d time.Duration) Renderer {
```

This kind of makes it nice for autocompletion purposes.

log/model_formats.go
```diff
@@ -0,0 +1,57 @@
...
+	return &any{d.String()}
+}
+
+func Timestamp(t time.Time) Renderer {
```
vxio, line 43:
```suggestion
func Time(t time.Time) Renderer {
```

log/model_formats.go
```diff
@@ -0,0 +1,57 @@
...
+	return TimestampFormat(t, time.RFC3339Nano)
+}
+
+func TimestampFormat(t time.Time, format string) Renderer {
```
vxio, line 47:
```suggestion
func TimeFormatted(t time.Time, format string) Renderer {
```

log/model_formats.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"encoding/base64"
+	"time"
+)
+
+// Renderer is an interface to deal with typing problems of just having an interface{} as the acceptable parameters
+// Go-kit logging has a failure case if you attempt to throw any values into it.
+// This is a way to guard our developers from having to worry about error cases of the lower logging framework.
+type Renderer interface {
```
vxio, line 11: nit: should we call this `Valuer`?

vxio:
Could we add in method chaining? 

```go
// current
log = log.Debug()
log.Set("ts", log.Time(time.Now().UTC()))
log.Set("foo", log.String("bar"))

// version with method chaining
log.Debug().
      Time("ts", time.Now().UTC()).
      String("foo","bar")
     
```

log/model_formats.go
InfernoJJ: I think it makes sense that we're letting them know were casting it to a string.

adamdecaf (approved):

vxio (approved):

PR #140, opened 2021-01-13
stringOrNil valuer

vxio:
With this, we won't have to write something like:
```go
	if event.OrganizationID != nil {
		logger.Set("organizationID", log.String(*event.OrganizationID))
	} else {
		logger.Set("organizationID", log.String("null"))
	}
```


darwinz (approved):
Nice!

InfernoJJ (approved):

PR #159, opened 2021-04-09
Adding in a .Nil() method to the logged error

InfernoJJ:
Adding in a .Nil() method to the logged error so you can log and return nil in the same oneliner

### DIFF
--- a/log/logger_impl.go
+++ b/log/logger_impl.go
@@ -140,3 +140,7 @@ type LoggedError struct {
 func (l LoggedError) Err() error {
 	return l.err
 }
+
+func (l LoggedError) Nil() error {
+	return nil
+}


log/logger_impl.go
```diff
@@ -140,3 +140,7 @@ type LoggedError struct {
 func (l LoggedError) Err() error {
 	return l.err
 }
+
+func (l LoggedError) Nil() error {
```
vxio, line 144:
```suggestion
func (l LoggedError) Nil() interface{} {
```

vxio: Actually, `error` might be better here, up to you

vxio (approved):

PR #160, opened 2021-04-09
add subrouter function to Server

atonks2:
This PR adds a function to the `Server` struct that creates a path prefixed subrouter. This enables users to use different middlewares for different route groups.

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -101,6 +101,23 @@ func (s *Server) AddVersionHandler(version string) {
 	})
 }
 
+// Subrouter creates and returns a subrouter with the specific prefix.
+//
+// The returned subrouter can Use() middleware without impacting
+// the parent router. For example:
+//
+// 	  svr := NewServer(":8080")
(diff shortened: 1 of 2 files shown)


admin/admin.go
```diff
@@ -101,6 +101,23 @@ func (s *Server) AddVersionHandler(version string) {
 	})
 }
 
+// Subrouter creates and returns a subrouter with the specific prefix.
+//
+// The returned subrouter can Use() middleware without impacting
```
adamdecaf, line 106: I'm not sure what `can Use() middleware` refers to. Does this refer to our zero-trust middleware or something else?

atonks2: It's a method on mux.Router

adamdecaf (approved):

admin/admin.go
atonks2: I can get rid of the function reference and just say `use` instead of `Use()`

adamdecaf (approved):

PR #165, opened 2021-05-11
add tx signatures

vxio:


adamdecaf:
What's this for? It looks like something for moovfinanical apps and isn't a bugfix, so I'm confused why it was released as such. @vxio @InfernoJJ ?

PR #168, opened 2021-06-09
add error response wrapper type

hampgoodwin:
# Changes

- Add a error response wrapper type

# Why?

- To create a consistent user experience of errors across moov services

### DIFF
--- a/http/response.go
+++ b/http/response.go
@@ -21,6 +21,15 @@ var (
 	ErrNoUserID = errors.New("no X-User-Id header provided")
 )
 
+// ErrorResponse is a http response wrapper type for all moov errors. This type is used to give
+// a consistency user experience with errors.
+type ErrorResponse struct {
+	// Error describes the error
+	Error string `json:"error"`
+	// Data is any additional data used to enrich the Error description. For example, ozzo validation errors.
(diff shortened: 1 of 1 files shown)


nitharios:
Since this is a WIP, should it be a draft?

hampgoodwin:
> Since this is a WIP, should it be a draft?

Sure. I thought we all felt pretty good about the model based on the slack discussion; but if we have more to jam on, let's jam in draft mode.

scott-moov:
The most user friendly error message’s I’ve worked with looked something like this. As a UI engineer you could just pass through the `errorMessage` to the UI if you wanted, but if you wanted to write your own you could switch off the `errorCode`. Firebase uses a pattern like this. Here are some of their [common error codes](https://firebase.google.com/docs/reference/js/firebase.auth.Error). The data object could be optional and specific for the type of error.

```
{ 
	"errorCode": "transfer/invalid-payment-method", 
	"errorMessage": "The payment method you specified cannot perform this transfer", 
	"data": {
		"somekey": "somevalue"
	}
}
```

hampgoodwin:
Deleting since we decided we dont want this discussion in the open.

PR #171, opened 2021-06-21
database: mask mysql password in JSON marshaling

adamdecaf:
This is useful to hide passwords with `GET /config` on admin servers.

### DIFF
--- a/database/model_config.go
+++ b/database/model_config.go
@@ -1,6 +1,11 @@
 package database
 
-import "time"
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/moov-io/base/mask"
+)
 
 type DatabaseConfig struct {
 	MySQL        *MySQLConfig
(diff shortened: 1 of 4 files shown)


database/model_config.go
```diff
@@ -15,6 +18,21 @@ type MySQLConfig struct {
...
+	type Aux struct {
+		Address     string
+		User        string
+		Password    string
+		Connections ConnectionsConfig
+	}
+	return json.Marshal(Aux{
+		Address:     m.Address,
+		User:        m.User,
+		Password:    "a***z",
```
vxio, line 31: I think this needs to be `mask.Password(..)`

vxio (approved):

mask/password.go
```diff
@@ -0,0 +1,21 @@
...
+	"unicode/utf8"
+)
+
+func Password(s string) string {
+	if utf8.RuneCountInString(s) < 3 {
+		return "**" // too short, we can't mask anything
+	} else {
+		// turn 'password' into 'p******d'
+		first, last := s[0:1], s[len(s)-1:]
+		return fmt.Sprintf("%s%s%s", first, strings.Repeat("*", len(s)-2), last)
```
atonks2, line 19: Does it matter that this exposes the first and last chars, and the length of the password?

adamdecaf: This is designed for admin servers of applications and to give operators some level of visibility to ensure the password is correct without exposing the credentials. It's a tradeoff and currently applications would render the entire password.

atonks2: I wonder if we could use a fixed amount of `*`? That'd still be enough for the admin to recognize the first and last chars, but not leak the password length.

adamdecaf: Yea we can do that. 5 stars?

atonks2: I think that sounds good.

atonks2 (approved):

PR #172, opened 2021-06-25
Set meaningful names for databases used in tests

vxio:
It's really helpful for debugging and development when I can look into the test database. For our services, a new database is created for each test. Without a meaningful name, I can't determine which database relates to which test run. The goal of this PR is to add meaning to the database name for local development.


database/mysql.go
```diff
@@ -187,15 +186,21 @@ func CreateTestMySQLDB(t *testing.T) *TestMySQLDB {
...
 		return "", err
 	}
 	defer db.Close()
 
-	dbName := "test" + base.ID()
+	// Set dbName to something like "TestCreateTemporaryD.._Jun_24_20:24:46"
+	dbName := fmt.Sprintf(
+		"%s.._%s",
+		t.Name()[:20],
+		time.Now().Local().Format(time.Stamp),
```
adamdecaf, line 201: Are spaces allowed in database names? Why not a format like t.Name() + "YYYY-MM-DD-HHmmss" ?

adamdecaf: Right, but the format doesn't work.

```
Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '.._Jun_25_15:02:27' at line 1
```

adamdecaf (approved):

PR #178, opened 2021-07-20
unexpected behaviour AddBankingDay

nean:
PayGate Version: `0.10.3`

**What were you trying to do?**
Coming from Paygate, the Effective Entry Date of an ACH entry was not calculated properly over weekends and was caught by Bank Validations
related: https://github.com/moov-io/paygate/pull/644

**What did you expect to see?**
Given any friday
`t.AddBankingDay(2)` adds 2 banking days to given date (monday, tuesday) and should return tuesday

**What did you see?**
`t.AddBankingDay(2)` adds 2 days to given date and if the date is still a holiday it adds 1 banking day recursively
This means the code doesnt check if the first added day is banking day or not

**How can we reproduce the problem?**
added failing test cases

### DIFF
--- a/time.go
+++ b/time.go
@@ -118,13 +118,19 @@ func (t Time) IsBankingDay() bool {
 	return true
 }
 
-// AddBankingDay takes an integer for the number of valid banking days to add and returns a Time
+// AddBankingDay takes an integer for the number of valid banking days to add and returns a Time.
+// Negative values and large values (over 500 days) will not modify the Time.
 func (t Time) AddBankingDay(d int) Time {
-	t.Time = t.Time.AddDate(0, 0, d)
-	if !t.IsBankingDay() {
-		return t.AddBankingDay(1)
+	if d < 1 || d > 500 {
+		return t
 	}
-	return t
+
+	t.Time = t.Time.AddDate(0, 0, 1)
+	if t.IsBankingDay() {
+		return t.AddBankingDay(d - 1)
+	}
+
+	return t.AddBankingDay(d)
 }
 
 // IsWeekend reports whether the given date falls on a weekend.
--- a/time_test.go
+++ b/time_test.go
@@ -227,6 +227,7 @@ func TestTime__IsWeekend(t *testing.T) {
 }
 
 func TestTime_AddBankingDay(t *testing.T) {
+	unchangeable := time.Date(2021, time.July, 15, 1, 0, 0, 0, est)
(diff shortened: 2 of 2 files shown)


nean:
recursive fix

```go
func (t Time) AddBankingDay(d int) Time {
	if d < 1 {
		return t
	}

	t.Time = t.Time.AddDate(0, 0, 1)
	if t.IsBankingDay() {
		return t.AddBankingDay(d - 1)
	}

	return t.AddBankingDay(d)
}
```
returns same day for negative values though

time.go
```diff
@@ -118,13 +118,19 @@ func (t Time) IsBankingDay() bool {
...
 }
 
-// AddBankingDay takes an integer for the number of valid banking days to add and returns a Time
+// AddBankingDay takes an integer for the number of valid banking days to add and returns a Time.
+// Negative values and large values (over 500 days) will not modify the Time.
 func (t Time) AddBankingDay(d int) Time {
-	t.Time = t.Time.AddDate(0, 0, d)
-	if !t.IsBankingDay() {
-		return t.AddBankingDay(1)
+	if d < 1 || d > 500 {
```
InfernoJJ, line 124: Looks odd to check for? Looks good otherwise.

InfernoJJ (approved):

time.go
adamdecaf: Adding negative banking days doesn't seem like a valid use-case. The 500+ guard is just to prevent huge recursive calls. IIRC Go doesn't perform tail call optimization so passing large values could OOM the service.

nean: will it be better to raise an error instead of returning if its out of bounds ?

adamdecaf: Yea, but that would require changing and breaking the method signature.

nean: Nothing on my side then, it would be nice if you could tag a paygate version after this merge

adamdecaf (approved):

adamdecaf:
Thanks for the fix @nean! I've released a bugfix of paygate with the upgrade.  https://github.com/moov-io/paygate/releases/tag/v0.10.4

nean:
@adamdecaf just noticed that the new paygate image was pushed to the docker tag `v0.10.3` instead of `v0.10.4`
might be some miss in the release script ?

adamdecaf:
@nean Yep, that's my bad. I've started a fixed build of v0.10.4

PR #182, opened 2021-08-10
logger: add more int and float valuers

vxio:


adamdecaf (approved):

adamdecaf:
Do we have other logging changes to apply? This needs another minor release (v0.23.0) since it adds to the public API surface.

vxio:
@adamdecaf looks like no other changes for now. I'll cut this in a minor release, thanks!

PR #183, opened 2021-08-11
add WithPrefix for prefixed logging

hampgoodwin:
# Changes

- Add interface WithPrefix
- Implement logger.WithPrefix method, which users underlying go-kit/log.WithPrefix()
- Add test for new WithPrefix

# Why

I need to set prefix values at construction of a logger, or upon entering a a function set function-relevant values in the logger which will persist after serially logging. WithPrefix allows for setting values which will always be logged when .Log() is called, until that logger is cleaned up.

# Example use

```go
func (s *service) Get(id string, requestID string) (*Widget, error) {
contextualLogger := s.logger.WithPrefix(log.Fields{"function": log.String("service.Get"), "request_id": log.String(request_id)})

contextualLogger.Debug().Log("getting widget from store")
widget, err := s.store.Get(id)
if err !=nil {
  contextualLogger.Error().Log("something bad happened")
  return nil, err
}

contextualLogger.Debug().Log("returning widget")
return widget, nil
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v0.23.0 (Released 2021-08-11)
+
+IMPROVEMENTS
+
+- log: added WithPrefix method for persistence contextual logging
+
 ## v0.22.0 (Released 2021-08-09)
 
 IMPROVEMENTS
--- a/log/logger.go
+++ b/log/logger.go
@@ -2,6 +2,7 @@ package log
 
 type Logger interface {
 	Set(key string, value Valuer) Logger
+	WithPrefix(ctxs ...Context) Logger
 	With(ctxs ...Context) Logger
 
 	Debug() Logger
--- a/log/logger_impl.go
+++ b/log/logger_impl.go
@@ -51,6 +51,21 @@ func (l *logger) Set(key string, value Valuer) Logger {
 	})
 }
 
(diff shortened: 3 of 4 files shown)


hampgoodwin:
Hrmm, I didn't touch anything in the database/* dir, how am I bringing down the test coverage?

InfernoJJ (changes requested):
I don't understand the purpose of it? It sounds like just `.With` but mutates versus returning a thread safe new context?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## v0.23.0 (Released 2021-08-11)
+
+IMPROVEMENTS
+
+- log: added WithPrefix method for persistence contextual logging
```
adamdecaf, line 5: https://github.com/moov-io/base/pull/182 is part of this release too, right?

hampgoodwin:
Closing for now.
After discussion with @InfernoJJ we think there is a different issue that causes .With() to not work as intended on larger subsets of contexts. I'll work on reproducing in a test, and then I'll PR some fixes.

Ah,, #182 will be included the next release  unless someone beats me to it.

hampgoodwin:
Ok, some further digger and testing revealed the logger package .With works as intended. The issue was a cli log parser that I run which was de-duping values.

PR #192, opened 2021-09-24
MySQL TLS

alexjplant:
**NOTE**: In this configuration certificate verification (which in golang TLS lib parlance means hostname and cert chain checking both) is disabled and replaced with a custom method that still verifies the CA.

This time I tested this locally first :-). The MySQL official Docker image automatically generates server certs. I extracted the server CA using `docker run` and `cat`:

```
Certificate:                                                                                                         
    Data:                                                                                                            
        Version: 3 (0x2)                                                                                             
        Serial Number: 1 (0x1)                                                                                       
    Signature Algorithm: sha256WithRSAEncryption                                                                     
        Issuer: CN=MySQL_Server_8.0.26_Auto_Generated_CA_Certificate
        Validity                                                                                                     
            Not Before: Sep 23 22:26:41 2021 GMT                                                                     
            Not After : Sep 21 22:26:41 2031 GMT                                                                     
        Subject: CN=MySQL_Server_8.0.26_Auto_Generated_CA_Certificate
```

I then saved it to my local machine and tested starting the app using the following configuration:

```
  Database:
    DatabaseName: "bankaccounts"
    MySQL:
      Address: "tcp(localhost:3306)"
      User: "root"
      Password: "password"
      TLSCAFile: "/Users/alexjplant/scratch/ca.pem"
```

It was able to connect successfully (I just didn't bother to make the database):

`ts=2021-09-24T18:03:14Z msg="attempting to connect to database again: Error 1049: Unknown database 'bankaccounts'" version=v0.10.9 level=info app=bankaccounts errored=true`

### DIFF
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -94,25 +94,13 @@ func (my *mysql) Connect(ctx context.Context) (*sql.DB, error) {
 	return db, nil
 }
 
-func mysqlConnection(logger log.Logger, user, pass string, address string, database string, sslCA string) *mysql {
+func mysqlConnection(logger log.Logger, user, pass string, address string, database string) *mysql {
 	timeout := "30s"
 	if v := os.Getenv("MYSQL_TIMEOUT"); v != "" {
 		timeout = v
 	}
-
-	var params []string
-	params = append(params, fmt.Sprintf("timeout=%s", timeout))
-	params = append(params, "charset=utf8mb4")
-	params = append(params, "parseTime=true")
-	params = append(params, `sql_mode="ALLOW_INVALID_DATES,STRICT_ALL_TABLES"`)
-
-	if sslCA != "" {
-		params = append(params, fmt.Sprintf("sslca=%s", sslCA))
-	}
-
-	renderedParams := strings.Join(params, "&")
-
-	dsn := fmt.Sprintf("%s:%s@%s/%s?%s", user, pass, address, database, renderedParams)
+	params := fmt.Sprintf(`timeout=%s&charset=utf8mb4&parseTime=true&sql_mode="ALLOW_INVALID_DATES,STRICT_ALL_TABLES"`, timeout)
+	dsn := fmt.Sprintf("%s:%s@%s/%s?%s", user, pass, address, database, params)
 	return &mysql{
 		dsn:         dsn,
 		logger:      logger,
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -5,9 +5,13 @@ package database
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"database/sql"
+	"encoding/pem"
 	"errors"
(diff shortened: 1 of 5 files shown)


database/mysql.go
```diff
@@ -94,30 +99,57 @@ func (my *mysql) Connect(ctx context.Context) (*sql.DB, error) {
...
-	params = append(params, "charset=utf8mb4")
-	params = append(params, "parseTime=true")
-	params = append(params, `sql_mode="ALLOW_INVALID_DATES,STRICT_ALL_TABLES"`)
+	var tlsConfig *tls.Config = nil
 
-	if sslCA != "" {
-		params = append(params, fmt.Sprintf("sslca=%s", sslCA))
+	if tlsCAFile != "" {
+		tlsConfig = &tls.Config{
+			InsecureSkipVerify: true,
```
adamdecaf, line 113: This seems like something we should accept on the `DatabaseConfig` rather than hardcode deep in our library.

alexjplant: I figured that since we were doing the CA check this covered that. I'll create a config option for this and for our CA-only check so that configurations can choose which combination to enforce.

database/mysql.go
```diff
@@ -94,30 +99,69 @@ func (my *mysql) Connect(ctx context.Context) (*sql.DB, error) {
...
+				}
+
+				if appendOK := rootCertPool.AppendCertsFromPEM(certPem); !appendOK {
+					return nil, err
+				}
+				tlsConfig.RootCAs = rootCertPool
+
+				if mysqlConfig.VerifyCAFile {
+					tlsConfig.VerifyConnection = func(state tls.ConnectionState) error {
+						if !(state.PeerCertificates[len(state.PeerCertificates)-1].Equal(caCert)) {
```
adamdecaf, line 139: We should perform a length check here, right? This could panic if the chain has one certificate.

alexjplant: It'd panic if it had 0 certificates (length 1 - 1 = 0 which is a one-element slice) and [the docs](https://pkg.go.dev/crypto/tls) state

> // On the client side, it can't be empty.

database/mysql.go
```diff
@@ -94,30 +99,69 @@ func (my *mysql) Connect(ctx context.Context) (*sql.DB, error) {
...
+				}
+
+				block, _ := pem.Decode(certPem)
+				caCert, err := x509.ParseCertificate(block.Bytes)
+				if err != nil {
+					return nil, err
+				}
+
+				if appendOK := rootCertPool.AppendCertsFromPEM(certPem); !appendOK {
+					return nil, err
```
adamdecaf, line 133: Shouldn't this return something like `errors.New("unable to append certificate")`?

adamdecaf (approved):

adamdecaf (approved):

PR #201, opened 2021-11-04
config: fail loading if there are unused (extra) fields

adamdecaf:
Often developers will accidently typo values in YAML files which are hard to spot. Thankfully mapstructure offers a way to error in this case which presents the error to developers quickly.


vxio (dismissed):
nice check!

config/config.go
```diff
@@ -54,7 +61,7 @@ func (s *Service) LoadFile(file string, config interface{}) error {
 		return logger.LogErrorf("unable to load the defaults: %w", err).Err()
 	}
 
-	if err := deflt.Unmarshal(config); err != nil {
+	if err := deflt.Unmarshal(config, errorOnUnusedFields); err != nil {
```
alovak, line 64: Is `UnmarshalExact` has the same effect?

adamdecaf: Probably? It loads some other functions so I wasn't sure without checking all the properties.

alovak (approved):

InfernoJJ (approved):

PR #210, opened 2022-01-11
Adding in additional metrics to track for database connections

InfernoJJ:
Adding in additional metrics to track for database connections

### DIFF
--- a/database/mysql.go
+++ b/database/mysql.go
@@ -38,6 +38,17 @@ var (
 		Help: "How many MySQL connections and what status they're in.",
 	}, []string{"state"})
 
+	mysqlConnectionsCounters = kitprom.NewGaugeFrom(stdprom.GaugeOpts{
+		Name: "mysql_connections_counters",
+		Help: `Counters specific to the sql connections. 
+			wait_count: The total number of connections waited for.
+			wait_duration: The total time blocked waiting for a new connection.
+			max_idle_closed: The total number of connections closed due to SetMaxIdleConns.
+			max_idle_time_closed: The total number of connections closed due to SetConnMaxIdleTime.
+			max_lifetime_closed: The total number of connections closed due to SetConnMaxLifetime.
+		`,
+	}, []string{"counter"})
+
 	// mySQLErrDuplicateKey is the error code for duplicate entries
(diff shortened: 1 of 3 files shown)


database/mysql.go
```diff
@@ -92,6 +104,11 @@ func (my *mysql) Connect(ctx context.Context) (*sql.DB, error) {
 				my.connections.With("state", "idle").Set(float64(stats.Idle))
 				my.connections.With("state", "inuse").Set(float64(stats.InUse))
 				my.connections.With("state", "open").Set(float64(stats.OpenConnections))
+				my.counters.With("stat", "wait_count").Set(float64(stats.WaitCount))
```
adamdecaf, line 107: This is called `counter` in the metric definition.

InfernoJJ: Yup, just caught that myself too.

database/mysql.go
```diff
@@ -38,6 +38,17 @@ var (
 		Help: "How many MySQL connections and what status they're in.",
 	}, []string{"state"})
 
+	mysqlConnectionsCounters = kitprom.NewGaugeFrom(stdprom.GaugeOpts{
+		Name: "mysql_connections_counters",
```
adamdecaf, line 42: Shouldn't this be `_gauges` ?

InfernoJJ: I was going off the naming much like how the status one is going off of it. 
These are counter under the hood but the API for prometheus go doesn't allow me to _set_ a value on a counter.

adamdecaf: Right, gauges record the value at lots of points in time. Counters only increment. https://prometheus.io/docs/concepts/metric_types/#gauge

Prometheus namings try and refer some type-level information about the data. `_seconds` or `_count`

InfernoJJ: yeah hence the counter metric is labelled `wait_ms`

InfernoJJ:
```			wait_count: The total number of connections waited for.
			wait_duration: The total time blocked waiting for a new connection.
			max_idle_closed: The total number of connections closed due to SetMaxIdleConns.
			max_idle_time_closed: The total number of connections closed due to SetConnMaxIdleTime.
			max_lifetime_closed: The total number of connections closed due to SetConnMaxLifetime.```

adamdecaf (approved):

PR #213, opened 2022-01-18
Adding slack package for basic info/critical messaging

pa3ng:


adamdecaf:
Why does this need to be in moov-io/base?

pa3ng:
Hey @adamdecaf, me and @darwinz noticed we were duplicating slack messaging implementations in a couple of our services. We thought we could probably move the shared logic in the base package. What are your thoughts?

adamdecaf:
I've been trying to keep moov-io/base pretty minimal and contain packages which *all* Moov projects use. If we wanted to make a more specific package we could follow the route of https://github.com/moov-io/go-sftp. 

That said this package looks to be one file and ~2 constants so I think copying the code makes more sense than a shared package. 

From the [Go Proverbs](https://go-proverbs.github.io/), "A little copying is better than a little dependency."

pa3ng:
Thanks for your input @adamdecaf. That makes sense to me; I'm cool w/ copypasta for now until we see more projects needing it, since it's only a couple of projects atm 👍🏼

PR #234, opened 2022-06-14
feat: Add Juneteenth holiday to Time calendar

mhargrove:
Add Juneteenth to the list of standard national holidays as per https://www.federalreserve.gov/aboutthefed/k8.htm

(this can be reverted if https://github.com/rickar/cal/pull/93 is merged & released)


time.go
```diff
@@ -43,7 +43,9 @@ func Now(location *time.Location) Time {
 	calendar := &cal.Calendar{
 		Name: "moov-io/base",
 	}
-	calendar.AddHoliday(us.Holidays...) // TODO(adam): check for more? June 19th?
+	holidays := us.Holidays
+	holidays = append(holidays, us.Juneteenth) // TODO: revert to just us.Holidays after https://github.com/rickar/cal/pull/93
```
darwinz, line 47: It looks like Juneteenth falls on a Sunday this year, and the following Monday (June 20) will be observed as a banking holiday.  We should add a test case for June 20th

adamdecaf:
I thought this was already handled by the library. I added it to the cal library before.

time.go
adamdecaf: Banking holidays are observed during the week, but businesses don't take Monday off when the holiday is on Sunday.

mhargrove: additional test case added in 200d7cd

darwinz (approved):

PR #252, opened 2022-09-08
database: add client cert to tls config

darwinz:


database/tls.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/moov-io/base/log"
+)
+
+const SQL_CLIENT_TLS_CERT = "SQL_CLIENT_TLS_CERT"
+const SQL_CLIENT_TLS_PRIVATE_KEY = "SQL_CLIENT_TLS_PRIVATE_KEY"
+
+func LoadTLSClientCertFromFile(logger log.Logger, certFile, keyFile string) (*tls.Certificate, error) {
+	if certFile == "" || keyFile == "" {
+		return nil, logger.Info().Logf("cert path or key path not provided").Err()
```
adamdecaf, line 16:
```suggestion
		return nil, logger.Error().Logf("cert path or key path not provided").Err()
```

database/mysql.go
```diff
@@ -168,6 +173,8 @@ func mysqlConnection(logger log.Logger, mysqlConfig *MySQLConfig, databaseName s
 						return nil
 					}
 				}
+
+				tlsConfig.Certificates = []tls.Certificate{&clientCert}
```
adamdecaf, line 177: What if we add that cert when it's not nil?
```suggestion
				if clientCert != nil {
					tlsConfig.Certificates = append(tlsConfig.Certificates, *clientCert(
				}
```

darwinz: I changed the `LoadTLSClientCertFromEnv` function to instead return a non-pointer Certificate.  I figure since I'm checking for the error, it should handle the error case.  But let me know if you think it's better to use a pointer and check for nil as you suggested.

InfernoJJ (commented):
Is it possible to get these values added to the config object? Environment variables would limit us to just one where the config would allow us for each config.

darwinz:
I didn't want to have to update all of our services.  Instead, terraform will create the client cert/key and include it in the deployment as an environment variable

adamdecaf:
We could add it to the struct and let an env var override those values.

database/tls_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test_LoadClientCertsFromConfig(t *testing.T) {
+
+	config := &database.MySQLConfig{
+		TLSClientCerts: []database.TLSClientCertConfig{
+			{
+				CertFilePath: "testdata/client_cert.pem",
+				KeyFilePath:  "testdata/client_cert_private_key.pem",
```
adamdecaf, line 17: These are missing, but can we build the paths in an OS-agnostic way?

```suggestion
				CertFilePath: filepath.Join("testdata", "client_cert.pem"),
				KeyFilePath:  filepath.Join("testdata", "client_cert_private_key.pem"),
```

gitleaks.toml
```diff
@@ -0,0 +1,7 @@
+title = "Gitleaks config"
+
+[rules.allowlist]
```
adamdecaf, line 3: What if we generated the certificate/key during the test? It'd save us from another config file in our repos. 

We use https://github.com/madflojo/testcerts elsewhere

adamdecaf (approved):

adamdecaf:
@darwinz Feel free to merge this and update the changelog for a v0.35.0 release.

PR #265, opened 2022-12-02
Fix MySQLUniqueViolation check to look for error dupe code more broadly

joshleckbee:
# Change
- Fixes MySQLUniqueViolation string matching to be less strict.

# Why
As of go-sql-driver mysql v1.7.0 the MySQL error code comes back as `Error {code} ({state}): {error message}`. Due to strict string matching the `MySQLUniqueViolation` func would fail to match since it looks for `Error {code}: {error message}`.


database/mysql.go
```diff
@@ -199,7 +199,7 @@ func mysqlConnection(logger log.Logger, mysqlConfig *MySQLConfig, databaseName s
 // MySQLUniqueViolation returns true when the provided error matches the MySQL code
 // for duplicate entries (violating a unique table constraint).
 func MySQLUniqueViolation(err error) bool {
-	match := strings.Contains(err.Error(), fmt.Sprintf("Error %d: Duplicate entry", mySQLErrDuplicateKey))
+	match := strings.Contains(err.Error(), fmt.Sprintf("Error %d", mySQLErrDuplicateKey))
```
mhargrove, line 202: can we do the same for `mysqlErrDataTooLong` so it's less fragile too?

kolaworld, line 202: Worth doing a similar change for method `MySQLDataTooLong`?

database/mysql.go
```diff
@@ -199,7 +199,7 @@ func mysqlConnection(logger log.Logger, mysqlConfig *MySQLConfig, databaseName s
...
 	if e, ok := err.(*gomysql.MySQLError); ok {
 		return match || e.Number == mySQLErrDuplicateKey
```
mhargrove, line 204: looks like [this](https://github.com/moov-io/base/blob/59aa06a665cd185868b743661723ef03d3857191/database/mysql.go#L47) is a dead link now, I think the new one would be https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_dup_entry if we want to go ahead and update it while we're in here

database/mysql_test.go
```diff
@@ -56,6 +56,13 @@ func TestMySQLUniqueViolation(t *testing.T) {
 	}
 }
 
+func TestMySQLUniqueViolation_WithStateValue(t *testing.T) {
+	err := errors.New(`problem upserting depository="282f6ffcd9ba5b029afbf2b739ee826e22d9df3b", userId="f25f48968da47ef1adb5b6531a1c2197295678ce": Error 1062 (23000): Duplicate entry '282f6ffcd9ba5b029afbf2b739ee826e22d9df3b' for key 'PRIMARY'`)
+	if !UniqueViolation(err) {
+		t.Error("should have matched unique violation")
+	}
+}
```
kolaworld, line 64: Following earlier comment, worth adding test method `TestMySQLDataTooLong_WithStateValue`?

database/mysql.go
mhargrove: (@kolaworld beat me to it - resolving this as [dupe](https://github.com/moov-io/base/pull/265#discussion_r1038576034))

vxio (dismissed):

database/mysql.go
joshleckbee: Done

database/mysql.go
joshleckbee: Done

database/mysql_test.go
joshleckbee: Done

database/mysql.go
joshleckbee: Done

kolaworld (approved):

mhargrove (approved):

InfernoJJ (approved):

PR #272, opened 2023-01-26
Increase max skip count to MaxInt64

pa3ng:


http/server.go
```diff
@@ -129,7 +129,7 @@ func GetSkipAndCount(r *http.Request) (skip int, count int, exists bool, err err
 		return skip, count, exists, err
 	}
 	// Cap skip at 10000
-	skip = int(math.Min(float64(skip), 10000))
+	skip = int(math.Min(float64(skip), math.MaxInt64))
```
adamdecaf, line 132: I'm not sure it's safe to increase this to 4 billion billion. What's the use-case where you'd skip that many rows? It seems a different design would be better in that case. Could we limit it to math.MaxInt32 or a billion?

adamdecaf:
```
$ go doc math.MaxInt64  | grep Int64
	MaxInt64  = 1<<63 - 1
```

pa3ng: Moved convo to slack

adamdecaf (approved):

PR #266, opened 2022-12-02
Update CHANGELOG.md

joshleckbee:

PR #287, opened 2023-05-01
add similar unique violation check as sql for spanner

gyao852:
Changes:
* Similar to what we support for `mysql::MySQLUniqueViolation`, added `spanner::SpannerUniqueViolation` that look at the error code or message to deduce if we have a duplicate error. 

Note: `spanner.Error` will be deprecated in the future, replaced with `APIError`. I added tests for both responses from client.
 
Screenshot 2023-05-01 at 4 21 25 PM

### DIFF
--- a/database/spanner.go
+++ b/database/spanner.go
@@ -2,14 +2,11 @@ package database
 
 import (
 	"database/sql"
-	"errors"
 	"fmt"
-	"strings"
 
 	"cloud.google.com/go/spanner"
 	"github.com/golang-migrate/migrate/v4/database"
 	migspanner "github.com/golang-migrate/migrate/v4/database/spanner"
-	"github.com/googleapis/gax-go/v2/apierror"
 	_ "github.com/googleapis/go-sql-spanner"
 	"google.golang.org/grpc/codes"
 
@@ -35,11 +32,5 @@ func SpannerMigrationDriver(cfg SpannerConfig, databaseName string) (database.Dr
 // Refer to https://cloud.google.com/spanner/docs/error-codes for Spanner error definitions,
 // and https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto for error codes
 func SpannerUniqueViolation(err error) bool {
-	match := strings.Contains(err.Error(), "Failed to insert row with primary key")
-
-	var apiErr *apierror.APIError
-	if errors.As(err, &apiErr) {
-		return match || spanner.ErrCode(apiErr) == codes.AlreadyExists
-	}
-	return match
+	return spanner.ErrCode(err) == codes.AlreadyExists
 }
--- a/database/spanner.go
+++ b/database/spanner.go
@@ -3,6 +3,7 @@ package database
 import (
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"cloud.google.com/go/spanner"
 	"github.com/golang-migrate/migrate/v4/database"
@@ -32,5 +33,6 @@ func SpannerMigrationDriver(cfg SpannerConfig, databaseName string) (database.Dr
 // Refer to https://cloud.google.com/spanner/docs/error-codes for Spanner error definitions,
 // and https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto for error codes
 func SpannerUniqueViolation(err error) bool {
-	return spanner.ErrCode(err) == codes.AlreadyExists
+	match := strings.Contains(err.Error(), "Failed to insert row with primary key")
+	return match || spanner.ErrCode(err) == codes.AlreadyExists
 }
--- a/database/spanner.go
+++ b/database/spanner.go
@@ -2,13 +2,18 @@ package database
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
+	"strings"
 
-	"github.com/moov-io/base/log"
-
+	"cloud.google.com/go/spanner"
 	"github.com/golang-migrate/migrate/v4/database"
 	migspanner "github.com/golang-migrate/migrate/v4/database/spanner"
+	"github.com/googleapis/gax-go/v2/apierror"
 	_ "github.com/googleapis/go-sql-spanner"
+	"google.golang.org/grpc/codes"
+
+	"github.com/moov-io/base/log"
 )
 
 func spannerConnection(logger log.Logger, cfg SpannerConfig, databaseName string) (*sql.DB, error) {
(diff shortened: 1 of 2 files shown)


database/spanner.go
```diff
@@ -24,3 +29,17 @@ func SpannerMigrationDriver(cfg SpannerConfig, databaseName string) (database.Dr
...
+// SpannerUniqueViolation returns true when the provided error matches the Spanner code
+// for duplicate entries (violating a unique table constraint).
+// Refer to https://cloud.google.com/spanner/docs/error-codes for Spanner error definitions,
+// and https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto for error codes
+func SpannerUniqueViolation(err error) bool {
+	match := strings.Contains(err.Error(), "Failed to insert row with primary key")
+
+	var apiErr *apierror.APIError
+	if errors.As(err, &apiErr) {
+		return match || spanner.ErrCode(apiErr) == codes.AlreadyExists
```
vxio, line 42: I think you can remove `match` in this case if you're already doing `spanner.ErrCode`

vxio (changes requested):

database/spanner.go
gyao852: In the [example](https://gist.github.com/gyao852/0319adb1c5bb521194147a8ff23b3769) they recommended to cast to `APIError` to get full information (like `Desc` and `additionalInformation` I think) but yeah I guess if we only need the `ErrCode` part I can just unwrap directly.

database/spanner.go
```diff
@@ -24,3 +29,17 @@ func SpannerMigrationDriver(cfg SpannerConfig, databaseName string) (database.Dr
...
 }
+
+// SpannerUniqueViolation returns true when the provided error matches the Spanner code
+// for duplicate entries (violating a unique table constraint).
+// Refer to https://cloud.google.com/spanner/docs/error-codes for Spanner error definitions,
+// and https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto for error codes
+func SpannerUniqueViolation(err error) bool {
+	match := strings.Contains(err.Error(), "Failed to insert row with primary key")
+
+	var apiErr *apierror.APIError
```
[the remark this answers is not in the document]
gyao852: In the [example](https://gist.github.com/gyao852/0319adb1c5bb521194147a8ff23b3769) they recommended to cast to `APIError` to get full information (like `Desc` and `additionalInformation` I think) but yeah I guess if we only need the `ErrCode` part I can just unwrap directly.

database/spanner.go
gyao852: done

adamdecaf (dismissed):

adamdecaf:
Are we going to close https://github.com/moov-io/base/pull/286 ? 

This can be released as v0.41.0 since it adds to the public API. Please commit the changelog update in a standalone commit and `git tag` it. You can create a release from the tag with the changelog update.

adamdecaf:
> Sounds good @adamdecaf ; I'm guessing you're referring to something like this #266 for the `changelog` commit? If so just need another stamp from Vince when he starts work and I'll have it in before EOD.

Yea but just commit the changelog to `master` and tag it.

database/spanner_test.go
```diff
@@ -69,3 +74,19 @@ func Test_MigrateAndRun(t *testing.T) {
...
+func TestSpannerUniqueViolation(t *testing.T) {
+	errMsg := "Failed to insert row with primary key ({pk#primary_key:\"282f6ffcd9ba5b029afbf2b739ee826e22d9df3b\"}) due to previously existing row"
+	// Test backwards-compatible parsing of spanner.Error (soon to be deprecated) from Spanner client
+	oldSpannerErr := spanner.ToSpannerError(status.New(codes.AlreadyExists, errMsg).Err())
+	if !database.SpannerUniqueViolation(oldSpannerErr) {
+		t.Error("should have matched unique violation")
+	}
+
+	// Test new apirerror.APIError response from Spanner client
+	newSpannerErr, parseErr := apierror.FromError(status.New(codes.AlreadyExists, errMsg).Err())
```
vxio, line 87: nit: you can dedupe `status.New(codes.AlreadyExists, errMsg)`

database/spanner_test.go
```diff
@@ -69,3 +74,19 @@ func Test_MigrateAndRun(t *testing.T) {
...
+	require.True(t, parseErr)
+	if !database.SpannerUniqueViolation(newSpannerErr) {
+		t.Error("should have matched unique violation")
+	}
```
vxio, line 91: @gyao852 could you add one more test case checking the unique violation is still caught if we wrap the error with `fmt.Errorf("wrapped err: %w")`

vxio (approved):
approved w/ two minor comments

database/spanner_test.go
gyao852: Ah, so if we wrap the error, the spanner library doesn't unwrap it since it ideally assumes it should be one of the two client responses above: 
Screenshot 2023-05-03 at 6 01 16 PM

When would we expect the error to be wrapped? Because I thought the repo layer will first catch this, and then we flag it ourselves before passing that upstream?

Regardless I re-added the `match` statement since I **think** all `err.Error()` will contain the base description and thus, Google's `Failed to insert row with primary key` string:
Screenshot 2023-05-03 at 6 05 06 PM

database/spanner_test.go
gyao852: done

database/spanner.go
```diff
@@ -24,3 +27,12 @@ func SpannerMigrationDriver(cfg SpannerConfig, databaseName string) (database.Dr
...
 	return s.Open(fmt.Sprintf("spanner://projects/%s/instances/%s/databases/%s?x-migrations-table=spanner_schema_migrations&x-clean-statements=true", cfg.Project, cfg.Instance, databaseName))
 }
+
+// SpannerUniqueViolation returns true when the provided error matches the Spanner code
+// for duplicate entries (violating a unique table constraint).
+// Refer to https://cloud.google.com/spanner/docs/error-codes for Spanner error definitions,
+// and https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto for error codes
+func SpannerUniqueViolation(err error) bool {
+	match := strings.Contains(err.Error(), "Failed to insert row with primary key")
+	return match || spanner.ErrCode(err) == codes.AlreadyExists
```
vxio, line 37:
```suggestion
	return spanner.ErrCode(err) == codes.AlreadyExists ||
		strings.Contains(err.Error(), "AlreadyExists")

```

database/spanner_test.go
vxio: Oh, looks like this is still an open issue: https://github.com/grpc/grpc-go/issues/2934

vxio (approved):

database/spanner.go
gyao852: done

adamdecaf (approved):

adamdecaf:
I squashed the commits since there was some cleanup. Do we plan on adding more before cutting a release?

gyao852:
While we could go down the rabbit hole of mapping all the errors, I think at least for now what we have is sufficient because we’re trying to replicate what we have existing for mySQL; there is the other sql verification `MySQLDataTooLong` which isn’t used heavily in our codebase nor does GCP Spanner docs mention any error codes mapped to this type of sql error. cc @vxio

vxio:
@gyao852 oh right, let's update that method (database::UniqueViolations), i didn't realize that existed.. thanks!

adamdecaf:
Sounds good. Let's get that patch in and can cut a release.

PR #302, opened 2023-05-31
Go:Embed

InfernoJJ:
Adds in support for go:embedding without breaking compatibility for pkger which is still used as the default for both migrations and default config files.

With adding it it adds in the options/curried functions pattern to be able to add configuration later, like the timeout one I added.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,6 +1,9 @@
 package config
 
 import (
+	"fmt"
+	"io"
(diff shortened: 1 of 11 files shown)


adamdecaf (commented):
LGTM aside from the comment.

adamdecaf (approved):

PR #311, opened 2023-07-05
Adding in the ability to specify with env variables how to format the log lines

InfernoJJ:
Adding in the ability to specify with env variables how to format the log lines

### DIFF
--- a/log/logger_impl.go
+++ b/log/logger_impl.go
@@ -13,11 +13,24 @@ import (
 )
 
 func NewDefaultLogger() Logger {
-	return NewLogger(log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)))
+	switch strings.ToLower(os.Getenv("MOOV_LOG_FORMAT")) {
+	case "json":
+		return NewJSONLogger()
+	case "nop", "noop":
+		return NewNopLogger()
+	case "logfmt":
+		return NewLogFmtLogger()
+	default:
+		return NewLogFmtLogger()
+	}
 }
 
 func NewNopLogger() Logger {
 	return NewLogger(log.NewNopLogger())
+}
+
+func NewLogFmtLogger() Logger {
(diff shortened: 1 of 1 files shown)


sarumont (approved):

log/logger_impl.go
```diff
@@ -13,13 +13,26 @@ import (
 )
 
 func NewDefaultLogger() Logger {
-	return NewLogger(log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)))
+	switch strings.ToLower(os.Getenv("MOOV_LOG_FORMAT")) {
```
adamdecaf, line 16: Do we need a Moov prefix?

InfernoJJ: It is a MOOV library and wouldn't want to interfere with anyone else using say a LOG_FORMAT value that doesn't match what we expect.

alovak: So, the default logger is not a default logger and it can be one of?

On one hand, I think it's an easy and quick solution with no code changes required. On the other hand, it changes the behavior of the "Default" logger. It can be a breaking change for tests.

We don't have a lot of places where we instantiate a logger in our projects (`environment.go`, and maybe some other). Maybe we can create a new method and update `templater` instead?

alovak:
So, the default logger is not a default logger and it can be one of?

On one hand, I think it's an easy and quick solution with no code changes required. On the other hand, it changes the behavior of the "Default" logger. It can be a breaking change for tests.

We don't have a lot of places where we instantiate a logger in our projects (environment.go, and maybe some other). Maybe we can create a new method and update templater instead?

InfernoJJ:
It doesn't change or break anything right now because no one uses that flag and its default was the same as before.

InfernoJJ:
The templater runs with NewDefault so it can use simply the logfmt or if this is set in the environment json. Most logging frameworks have a way without code to change the logging format, we currently don't.

adamdecaf (approved):

PR #314, opened 2023-07-21
log: make NewBufferLogger safe to read/write across goroutines

adamdecaf:


log/logger_test.go
```diff
@@ -297,7 +298,27 @@ func Test_Caller(t *testing.T) {
...
+	var wg sync.WaitGroup
+	wg.Add(1000)
+	for i := 0; i < 1000; i++ {
+		go func(idx int) {
+			defer wg.Done()
+
+			if idx%3 == 0 {
+				t.Log(buffer.String()) // read
+			} else {
+				log.Info().Logf("idx %d", idx)
```
gyao852, line 313: Why log it?

gyao852 (approved):

log/logger_test.go
adamdecaf: We want to make sure and call .Write to get into the `mu.Lock()` path of the logger.

PR #324, opened 2023-09-25
add orderBy param

michaelabbott25:
adding the `orderBy` param that way it's reusable and consistent if/when we add new endpoints that use ordering. this is [similar to how we handle paging](https://github.com/moov-io/base/blob/master/http/server.go#L122).


InfernoJJ:
I would get rid of the semi-confusing `exists` boolean being returned. We make all the values lowercase kabob formatted. So `createdOn` to `created-on` etc... This helps our customers who are developing against the platform to avoid mistakes.

InfernoJJ (approved):

PR #333, opened 2023-10-26
Check nil on fmt stringer

kolaworld:
- Prevent panic when calling `log.Stringer(nil)`

### DIFF
--- a/log/model_valuer.go
+++ b/log/model_valuer.go
@@ -93,6 +93,9 @@ func ByteBase64(b []byte) Valuer {
 }
 
 func Stringer(s fmt.Stringer) Valuer {
+	if s == nil {
+		return &any{nil}
+	}
 	return &any{s.String()}
 }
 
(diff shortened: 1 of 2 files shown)


log/model_valuer.go
```diff
@@ -93,6 +94,10 @@ func ByteBase64(b []byte) Valuer {
 }
 
 func Stringer(s fmt.Stringer) Valuer {
+	if reflect.ValueOf(s).Kind() == reflect.Pointer && reflect.ValueOf(s).IsNil() {
+		return &any{nil}
+	}
```
adamdecaf, line 99: I'm surprised checking `s == nil` doesn't work. 

Can we get the reflect value only once? 

```go
if v := reflect.ValueOf(s); v.Kind() == reflect.Pointer && v.IsNil() {
	return &any{nil}
}
```

kolaworld: Yep same on surprise `s == nil` doesn't work.

And updated to get value of reflect once.

adamdecaf (approved):

PR #392, opened 2024-05-10
Testing known github issue

ashton-herrington:
https://github.com/golang/go/issues/45993#issuecomment-1427873850

Tested this by setting the following configs 
`MaxLifetime: 1000s
MaxIdleTime: 3s`

and following code 
```
stats := env.DB.Stats()
fmt.Println("Open connections", stats.OpenConnections, "Idle connections", stats.Idle)

fmt.Println("Sleep")
// Should be long enough to clear idle connection
time.Sleep(5 * time.Second)

stats = env.DB.Stats()
fmt.Println("Open connections", stats.OpenConnections, "Idle connections", stats.Idle)
```

Before change:

```
Open connections 1 Idle connections 1
Sleep
Open connections 1 Idle connections 1
```

After change:

```
Open connections 1 Idle connections 1
Sleep
Open connections 0 Idle connections 0
```

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -79,6 +79,7 @@ func ApplyConnectionsConfig(db *sql.DB, connections *ConnectionsConfig, logger l
 		db.SetMaxIdleConns(connections.MaxIdle)
 	}
 
+	// Due to a known issue https://github.com/golang/go/issues/45993#issuecomment-1427873850, maxIdleTime must be specified in order before MaxLifetime else it will not be honored
 	if connections.MaxIdleTime > 0 {
 		logger.Logf("setting SQL max idle time to %v", connections.MaxIdleTime)
 		db.SetConnMaxIdleTime(connections.MaxIdleTime)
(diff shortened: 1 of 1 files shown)


database/database.go
```diff
@@ -79,15 +79,15 @@ func ApplyConnectionsConfig(db *sql.DB, connections *ConnectionsConfig, logger l
...
-	}
-
 	if connections.MaxIdleTime > 0 {
 		logger.Logf("setting SQL max idle time to %v", connections.MaxIdleTime)
 		db.SetConnMaxIdleTime(connections.MaxIdleTime)
 	}
 
+	if connections.MaxLifetime > 0 {
+		logger.Logf("setting SQL max lifetime to %v", connections.MaxLifetime)
+		db.SetConnMaxLifetime(connections.MaxLifetime)
```
adamdecaf, line 89: Can we include a comment about how the order is critical? I never would have guessed the ordering would have made a difference. 

https://github.com/golang/go/issues/45993#issuecomment-1427873850

ashton-herrington: yah makes sense, will do

anna-moov (approved):
I think based on what we investigated, this should hopefully fix the issues we are seeing in card-orch-batch.

adamdecaf:
This has been released in https://github.com/moov-io/base/releases/tag/v0.49.3

PR #416, opened 2024-07-31
log: set level=error for .LogError and .LogErrorf

adamdecaf:
Previously you could get logs at info/debug/etc levels with `errored=true` that aren't `level=error`.


InfernoJJ (approved):

adamdecaf:
Talked this over. It's probably too much of a breaking change to make this many years later.

PR #436, opened 2024-09-27
WIP: Support PostgreSQL Databases

carsonab:
Adds support for PostgreSQL:
- PostgreSQL database connections with TLS
  - Local testing can be done against a `alloydbomni` container
  - Added `gencerts.sh` to create certs used in local server config and client tests to enable TLS tests
- AlloyDB connections using [alloybd-go-connector](https://github.com/GoogleCloudPlatform/alloydb-go-connector)
- Uses single database driver `pgx`

Local testing can be done against a `alloydbomni` container, which uses the same database engine as hosted AlloyDB, using a normal postgres connection. Connections with a alloydb-go-connector need a GCP AlloyDB instance to connect to. However both connection methods use the same database driver. This means our local testing of business logic and its sql queries use the same database driver and engine as higher environments, and only the connection method will differ between environments.

Tested with alloybd-go-connector successfully using hosted AlloyDB instance for both IAM and native db login.

TODOs:
- handle closing dialer from `alloydbconn.NewDialer`

NOTE:
- Tested client certs by manually editing the server's `pg_hba.conf` to require client certs with `hostssl all all all scram-sha-256 clientcert=verify-full` but not automated to do that


adamdecaf:
Do we want to start this new database work in moovfinancial rather than in here?

database/migrator.go
```diff
@@ -155,6 +156,29 @@ func getDriver(logger log.Logger, config DatabaseConfig, opts *migrateOptions) (
 				return nil, nil, err
 			}
 		}
+	} else if config.Postgres != nil {
+		if opts.source == nil {
+			src, err := NewPkgerSource("postgres", false)
```
adamdecaf, line 161: We should be able to remove pkger. We've been using go:embed for quite a while. Can leave it in this PR for now.

database/postgres.go
```diff
@@ -0,0 +1,139 @@
...
+		dialer = d
+		dsn = fmt.Sprintf(
+			// sslmode is disabled because the alloy db connection dialer will handle it
+			"user=%s password=%s dbname=%s sslmode=disable",
+			config.User, config.Password, databaseName,
+		)
+	}
+
+	// TODO
+	//cleanup := func() error { return d.Close() }
```
adamdecaf, line 113: Needed?

database/postgres_test.go
```diff
@@ -0,0 +1,110 @@
...
+	config := database.DatabaseConfig{
+		DatabaseName: "moov",
+		Postgres: &database.PostgresConfig{
+			Address:  "localhost:5432",
+			User:     "moov",
+			Password: "moov",
+			TLS: &database.PostgresTLSConfig{
+				CACertFile:     "../testcerts/root.crt",
+				ClientCertFile: "../testcerts/client.crt",
+				ClientKeyFile:  "../testcerts/client.key",
```
adamdecaf, line 48: Need to use filepath.Join

database/postgres.go
carsonab: Yea, listed this under the TODOs in the description. We need to call this when we're done with the connection

database/migrator.go
InfernoJJ: yeah I'll remove that here soon.

alovak:
@carsonab thanks for the update! Looks solid!

carsonab:
Recreated branch on moov-io/base directly with maintainer permissions, closing this PR in favor https://github.com/moov-io/base/pull/437

PR #437, opened 2024-09-28
WIP: Support PostgreSQL Databases

carsonab:
Adds support for PostgreSQL:
- PostgreSQL database connections with TLS
  - Local testing can be done against a `alloydbomni` container
  - Added `gencerts.sh` to create certs used in local server config and client tests to enable TLS tests
- AlloyDB connections using [alloybd-go-connector](https://github.com/GoogleCloudPlatform/alloydb-go-connector)
- Uses single database driver `pgx`

Local testing can be done against a `alloydbomni` container, which uses the same database engine as hosted AlloyDB, using a normal postgres connection. Connections with a alloydb-go-connector need a GCP AlloyDB instance to connect to. However both connection methods use the same database driver. This means our local testing of business logic and its sql queries use the same database driver and engine as higher environments, and only the connection method will differ between environments.

Tested with alloybd-go-connector successfully using hosted AlloyDB instance for both IAM and native db login.

TODOs:
- [handle closing dialer](https://github.com/moov-io/base/pull/437/files#diff-8d10d7d8c3019b9ebb006b808d1f77e443e7772dc0a519f454bf35c533f08dedR113) from `alloydbconn.NewDialer`
- fix CI runners (these tests pass locally on macOS with docker desktop)

NOTE:
- Tested client certs by manually editing the server's `pg_hba.conf` to require client certs with `hostssl all all all scram-sha-256 clientcert=verify-full` but not automated to do that


docker-compose.yml
```diff
@@ -28,5 +28,22 @@ services:
     networks:
       - intranet
 
+  alloydbomni:
+    image: google/alloydbomni
+    restart: always
+    ports:
+      - "5432:5432"
```
adamdecaf, line 35: The container doesn't seem to be running. Are there fixes needed to the container? 

```
       	            	connecting to postgres: connecting to database: failed to connect to `user=moov database=moov`:
        	            		127.0.0.1:5432 (localhost): dial error: dial tcp 127.0.0.1:5432: connect: connection refused
        	            		[::1]:5432 (localhost): dial error: dial tcp [::1]:5432: connect: connection refused
```

https://github.com/moov-io/base/actions/runs/11108406519/job/30861274433?pr=437

database/postgres.go
```diff
@@ -0,0 +1,145 @@
...
+	if config.Alloy.UseIAM {
+		d, err := alloydbconn.NewDialer(ctx, alloydbconn.WithIAMAuthN())
+		if err != nil {
+			return "", fmt.Errorf("creating alloydb dialer: %v", err)
+		}
+		dialer = d
+		dsn = fmt.Sprintf(
+			// sslmode is disabled because the alloy db connection dialer will handle it
+			// no password is used with IAM
+			"user=%s dbname=%s sslmode=disable",
```
adamdecaf, line 97: Is there a reason we format a DSN here but for vanilla postgres use their URI syntax?

docker-compose.yml
carsonab: Yea I noticed, not sure why yet. It works locally on macOS. The `alloydbomni` image is known to not work with with Orbstack (which is true locally), but the runner should be using docker anyway. Also tried with postgres image, which fails with same problem. Will need to keep debugging.

database/postgres.go
carsonab: Either should work. I just preferred the `key=value` syntax when not giving a server (since thats handled by the `alloydbconn` library when connecting to a hosted AlloyDB instance)

InfernoJJ (approved):

PR #448, opened 2025-04-04
catch psql deadlock err

kolaworld:
- Catch psql deadlock error and return in helper `database.DeadlockFound()`

### DIFF
--- a/database/postgres.go
+++ b/database/postgres.go
@@ -132,29 +132,32 @@ func getAlloyDBConnectorConnStr(ctx context.Context, config PostgresConfig, data
 	return connStr, nil
 }
 
+// PostgresUniqueViolation returns true when the provided error matches the Postgres code
+// for unique violation.
 func PostgresUniqueViolation(err error) bool {
 	if err == nil {
 		return false
 	}
-	match := strings.Contains(err.Error(), postgresErrUniqueViolation)
 
 	var pgError *pgconn.PgError
-	if errors.As(err, &pgError) {
-		return match || pgError.Code == postgresErrUniqueViolation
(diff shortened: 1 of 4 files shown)


mhargrove (approved):

database/postgres.go
```diff
@@ -135,11 +136,25 @@ func PostgresUniqueViolation(err error) bool {
...
 		return false
 	}
+	match := strings.Contains(err.Error(), postgresErrUniqueViolation)
+
 	var pgError *pgconn.PgError
 	if errors.As(err, &pgError) {
-		if pgError.Code == postgresErrUniqueViolation {
-			return true
-		}
+		return match || pgError.Code == postgresErrUniqueViolation
```
vxio, line 143: I think you can remove the `match ||` here -- cause that'll be covered by the `return match` below?

InfernoJJ (dismissed):

database/postgres.go
kolaworld: No, if you remove `match ||` then `pgError.Code == postgresErrUniqueViolation` would return if it evaluates to false.

That return could be too eager if `match` evaluates to true.

vxio: mm i see, was mainly wondering if we could remove the `match` out of the `errors.As(err, &pgError` block -- think you could simplify this to:

```go
if errors.As(err, &pgError) && 
   pgError.Code == postgresErrUniqueViolation {
   return true
}

return strings.Contains(err.Error(), postgresErrUniqueViolation)
```

kolaworld: updated and added more tests [564f288](https://github.com/moov-io/base/pull/448/commits/564f2880d8f4d4d5ad76530f1d68ccf63024380b)

vxio (approved):

PR #449, opened 2025-04-07
fix lint errors

kolaworld:
- ~Add https://github.com/ccoveille/go-safecast to fix int casting errors~
  - ~Impact may be slightly breaking for casting in telemetry span attributes.~
  - ~Specifically, if casting int fails, we fallback to setting the span attribute as a string using formatting rather than dropping the span attribute.~
  
  - ~**Pro** of fallback: is we don't drop the span attribute no matter what.~
  - ~**Con** of fallback: is querying span attributes may break given type change~
  - ~See [e18436c](https://github.com/moov-io/base/pull/449/commits/e18436c6a9f1ed01f514590e8afaaf2c2725032a)~

- Add `//nolint:gosec` to resolve lint errors for casting int

- Add `//nolint:sqlclosecheck` to resolve lint errors given the lint failures are around wrapping database/sql methods. i.e. we're not supposed to close those.


InfernoJJ (changes requested):
I don't like this as its making the observability code the main focus of the function due to the number of lines it'll take up and the repetetive nature. It also seems like we're upcasting an int to int64 which will fit and thus not ever trip the string version.

Seems like we enabled another buggy linter.

PR #452, opened 2025-04-21
Don't print Go type in ParseError messages

cjao-lead:
`ParseError`s currently print the internal Go type of the wrapped `error` when formatted as strings. In practice this seems to be superfluous and clutters the error messages with information primarily of interest to `moov-io` developers. The wrapped `error` already aptly describes why parsing failed.

Since no other Moov errors display Go types when formatted, this PR removes the Go type from the string representation of `ParseError`s.

### DIFF
--- a/error.go
+++ b/error.go
@@ -30,9 +30,9 @@ type ParseError struct {
 
 func (e ParseError) Error() string {
 	if e.Record == "" {
-		return fmt.Sprintf("line:%d %T %s", e.Line, e.Err, e.Err)
+		return fmt.Sprintf("line:%d %s", e.Line, e.Err)
 	}
-	return fmt.Sprintf("line:%d record:%s %T %s", e.Line, e.Record, e.Err, e.Err)
+	return fmt.Sprintf("line:%d record:%s %s", e.Line, e.Record, e.Err)
 }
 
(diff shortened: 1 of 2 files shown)


adamdecaf:
I'm not sure we can merge this as there's code that relies on the type being included. We don't tend to use this ParseError in new code, but changing this would break current code.

cjao-lead:
Understood. Would you consider allowing the error formatting to be configurable? See the latest commit, which defaults to the existing behavior but allows users to opt into the alternate error presentation [like so](https://github.com/moov-io/ach/compare/master...cjao-lead:ach:casey/configure-err-formatting)?

cjao-lead:
For context, I'm interested in exposing moov error messages to users who won't be conversant in Moov internals or even Go. While it seems possible to parse the error I get from `Reader.Read()` to strip out the Go type, suppressing the extraneous information at its source would be more elegant.

adamdecaf:
You can expose our error messages to your customers, but we don't recommend that. I'm hesitent to start accepting configuration options like this - as it increases our burden for future features.

cjao-lead:
No problem, I'll consider other approaches. Thanks for your feedback.

PR #457, opened 2025-05-01
Add struct context

alovak:
Add the `StructContext` function that allows to log struct fields automatically by using tags.

```go
import "github.com/moov-io/base/log"

// Define a struct with log tags
type User struct {
    ID       int    `log:"id"`
    Username string `log:"username"`
    Email    string `log:"email,omitempty"` // won't be logged if empty
    Address  Address `log:"address"` // nested struct must have log tag
    Hidden   string // no log tag, won't be logged
}

//...
// Log with struct context
logger.With(log.StructContext(user)).Info().Log("User logged in")
```

### DIFF
--- a/log/README.md
+++ b/log/README.md
@@ -1,0 +1,122 @@
+# Log Package
+
+The log package provides structured logging capabilities for Moov applications.
+
+## Usage
+
+### Basic Logging
+
+```go
+import "github.com/moov-io/base/log"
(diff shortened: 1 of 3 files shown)


sarumont (approved):

alovak:
Now we can see what goes to logs and what to traces/spans :)

PR #473, opened 2025-09-29
feat: add retryable DB operations

sarumont:
There are certain cases where a DB-level retry is desired, namely those related to temporary connectivity issues to the database itself. This PR adds a generic rate-limited retry function as well as extensions to `DB` for retryable queries.

### DIFF
--- a/ratex/ratelimit.go
+++ b/ratex/ratelimit.go
@@ -1,0 +1,148 @@
+package ratex
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+	"math/big"
+	"time"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/time/rate"
+
+	"github.com/moov-io/base/telemetry"
+)
+
+type RateLimitParams struct {
+	RateLimiter *rate.Limiter // can be nil to create a new rate limiter
+	TryCount    int           `otel:"try_count"`
+	MinDuration time.Duration `otel:"min_duration_n"`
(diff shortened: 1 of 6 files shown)


ratex/ratelimit.go
```diff
@@ -0,0 +1,148 @@
...
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/time/rate"
+
+	"github.com/moov-io/base/telemetry"
+)
+
+type RateLimitParams struct {
+	RateLimiter *rate.Limiter // can be nil to create a new rate limiter
+	TryCount    int           `otel:"try_count"`
+	MinDuration time.Duration `otel:"min_duration_n"`
```
adamdecaf, line 20:
```suggestion
	MinDuration time.Duration `otel:"min_duration_ns"`
```

ratex/ratelimit.go
```diff
@@ -0,0 +1,148 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/time/rate"
+
+	"github.com/moov-io/base/telemetry"
+)
+
+type RateLimitParams struct {
+	RateLimiter *rate.Limiter // can be nil to create a new rate limiter
+	TryCount    int           `otel:"try_count"`
```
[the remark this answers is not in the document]
sarumont: Documented and renamed to (hopefully) make it a bit more clear just from the var name.

ratex/ratelimit.go
adamdecaf: Right but we can set it on the span as ms, rather than as nanos.

PR #475, opened 2025-10-06
refactor!: clarify variable naming for attempts in retryable operations

sarumont:
Use "attempts" for MaxAttempts and AttemptNum to clarify that the count includes the initial try and subsequent retries - i.e. a MaxAttempts of 3 is 1 try with 2 possible retries.

BREAKING CHANGE: renames MaxRetries -> MaxAttempts, RetryAttempt -> AttemptNum

### DIFF
--- a/ratex/ratelimit.go
+++ b/ratex/ratelimit.go
@@ -18,16 +18,17 @@ import (
 type RateLimitParams struct {
 	RateLimiter *rate.Limiter // can be nil to create a new rate limiter
 
-	// RetryAttempt represents the current retry attempt, starting at 1. This will increment for each retry
-	RetryAttempt int
-	MinDuration  time.Duration
-	MaxDuration  time.Duration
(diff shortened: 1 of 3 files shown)


ratex/ratelimit.go
```diff
@@ -18,16 +18,17 @@ import (
 type RateLimitParams struct {
 	RateLimiter *rate.Limiter // can be nil to create a new rate limiter
 
-	// RetryAttempt represents the current retry attempt, starting at 1. This will increment for each retry
-	RetryAttempt int
-	MinDuration  time.Duration
-	MaxDuration  time.Duration
+	// AttemptNum represents the current attempt, starting at 1. This will increment for each (re)try
+	AttemptNum int
```
adamdecaf, line 22: We should use full words here, so `Attempt int` or `AttemptNumber int`

adamdecaf (approved):

PR #490, opened 2026-03-17
database: add failover-safe pool defaults and pgxpool health checks for Postgres/AlloyDB

stevemsmith:
## Summary

Switch Postgres/AlloyDB connections from `sql.Open("pgx")` to `pgxpool` with a `*sql.DB` wrapper, and configure **acquire-time** liveness checks so dead connections left by AlloyDB maintenance switchovers are not handed to the application.

## Problem

During AlloyDB maintenance switchovers, Go services fail because dead connections sit in the pool and get handed to the application. The error looks like:

```
failed to connect to `user=tumbler_sa database=tumbler`: 10.2.10.25:5432 (10.2.10.25): tls error: read tcp ...: read: connection reset by peer
```

Without this change, services continue returning 500s until the pool naturally cycles out the dead connections — which with default `database/sql` settings can take a long time.

### `postgres.go`
- Replace `sql.Open("pgx", connStr)` with `pgxpool.NewWithConfig()` + custom `openDBFromPool` (wraps `stdlib.GetPoolConnector`)
- Set `ShouldPing` to ping connections idle **> 200ms** before handing them to the caller. This is the mechanism that actually tests liveness at acquire time — `HealthCheckPeriod` (the background reaper) only evicts connections exceeding age thresholds and does **not** ping
- Set `PingTimeout` default to 1s (bounds hung peers; RST failures still fail fast)
- Apply `PostgresConfig.Connections` onto **pgxpool** via `ApplyPostgresPoolConfig` / `ResolvePostgresConnectionsConfig` (not `database/sql` setters)
- `MaxIdle` has no pgxpool equivalent: logged and ignored when set
- Wrap the connector so `db.Close()` closes the underlying pgxpool **and** AlloyDB dialer (fixes the previous dialer leak TODO; plain `stdlib.OpenDBFromPool` does not close the pool)
- Refactor `getAlloyDBConnectorConnStr()` → `buildAlloyDBPoolConfig()` returning `*pgxpool.Config` + dialer (no more `stdlib.RegisterConnConfig`)

### `database.go`
- Stop calling `ApplyConnectionsConfig` on the Postgres `*sql.DB` path — those setters do not configure pgxpool, and `SetMaxIdleConns(n>0)` breaks `OpenDBFromPool` by preventing connections from returning to the pool
- `ApplyPostgresConnectionsConfig` is kept only as a **deprecated** helper for accidental sql.DB tuning; prefer `PostgresConfig.Connections` or `ApplyPostgresPoolConfig`

### `model_config.go`
- `DefaultPostgresConnectionsConfig()` supplies zeros filled by `ResolvePostgresConnectionsConfig`:
  - **MaxOpen = 25**
  - **MaxLifetime = 5m**
  - **MaxIdleTime = 2m**
  - **MaxIdle** is not defaulted (unused on Postgres)

## Operator notes (runtime behavior change)

API remains `*sql.DB` (no downstream code changes required), but **unset `Connections` fields no longer mean unlimited**:

| Field | Old (`database/sql` + pgx) | New (pgxpool defaults) |
|-------|----------------------------|-------------------------|
| MaxOpen unset | unlimited | **25** |
| MaxLifetime unset | no limit | **5m** (pgxpool adds jitter) |
| MaxIdleTime unset | no limit | **2m** |
| MaxIdle | sql.DB idle cap | **ignored** on Postgres (logged) |

**Action for services:** set `PostgresConfig.Connections` explicitly if you need a larger pool or longer idle/lifetime (especially high-QPS or AlloyDB IAM where reconnect is costly). There is no sentinel for “unlimited” lifetime/idle under pgxpool.

Also do **not** call `db.SetMaxIdleConns(n>0)` on DBs from `database.New` for Postgres — keep idle management on pgxpool (`MaxIdleConns` must stay 0 on the sql.DB wrapper).

## Out of scope

Retry logic is intentionally deferred. Retrying in-flight queries on connection failure needs explicit transaction handling to avoid double-writes.

## Non-breaking (API)

- Return type of `New()` is still `*sql.DB`
- MySQL and Spanner paths are untouched
- No new external module dependencies (`pgxpool` / `stdlib` are part of `pgx/v5`)

Runtime pool defaults above **are** a behavior change for services that left `Connections` zeroed — see Operator notes.

## Test plan

- [x] `go build ./database/...` and `go vet ./database/...`
- [x] Unit tests for `ResolvePostgresConnectionsConfig`, `ApplyPostgresPoolConfig`, and `openDBFromPool` Close lifecycle (see follow-up commit / companion PR if not yet on this branch)
- [ ] Dogfood on ≥1 AlloyDB service under maintenance / connection drain
- [ ] Confirm CI green on PR head


adamdecaf (dismissed):

adamdecaf (dismissed):

adamdecaf:
Can we try this in some apps before merging?

adamdecaf:
I'll work with gosec/golangci-lint on [getting the linter fix released](https://github.com/securego/gosec/pull/1615#issuecomment-4077604530).

adamdecaf:
@stevemsmith can you rebase off the latest origin/master ?

database/postgres.go
```diff
@@ -164,3 +174,77 @@ func PostgresDeadlockFound(err error) bool {
...
+	}
+
+	// PostgreSQL error codes indicating the server is shutting down or unavailable
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) {
+		switch pgErr.Code {
+		case "57P01", "57P02", "57P03": // admin_shutdown, crash_shutdown, cannot_connect_now
+			return true
+		case "08000", "08001", "08003", "08004", "08006": // connection_exception class
+			return true
```
InfernoJJ, line 193: I want to see documentation on these error cases as being valid to retry. Unsure I trust an AI here with possible data corruption if its incorrect. Or just remove the retry stuff and put it into a different PR.

database/postgres.go
```diff
@@ -164,3 +174,77 @@ func PostgresDeadlockFound(err error) bool {
...
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		err = fn()
+		if err == nil {
+			return nil
+		}
+		if !IsRetryablePostgresError(err) {
+			return err
+		}
+		if attempt < maxAttempts-1 {
+			backoff := time.Duration(attempt+1) * 200 * time.Millisecond
```
InfernoJJ, line 241: Backoff is a bad idea in this case as you'll drastically increase the wait, along with it not being configurable. 200ms is an ETERNITY to a program, and is noticeable by a user. Theirs also no variance in this interface, so they will all slam at the same time.

InfernoJJ (changes requested):
Remove the retry stuff, otherwise looks good.

database/postgres.go
elovelan: Both Gemini and Claude agree that at least some of these are unsafe to retry.

Interestingly, they both mentioned `40001` (`serialization_failure`) and `40P01` (`deadlock_detected`) _are_ safe even though they're not included here. [The guide](https://www.postgresql.org/docs/current/mvcc-serialization-failure-handling.html) for them mentions that there are a few similar codes `23505` (`unique_violation`), and `23P01` (`exclusion_violation`) that can be safe to retry, but might need to be handled on a case-by-case basis based on the way the calling code is written.

`57P01` (`admin_shutdown`) is definitely safe based on [the docs](https://www.postgresql.org/docs/current/server-shutdown.html) because the transaction is rolled back during a fast shutdown. I can't find a definitive source that says that `57P03` (`cannot_connect_now`) is safe, but the name for the code, as well as non-authoritative sources, including Claude and Gemini, indicate it's safe to retry. Two others they mention are safe are `53300` (`too_many_connections`), which is a connect-time error, and `57014` (`query_canceled`) because it indicates a rollback.

Interestingly, even though `08001` (`sqlclient_unable_to_establish_sqlconnection`) and `08004` (`connection_rejected`) are safe, `080xx` codes are designed to be implemented by the client and `pgx` doesn't implement any, returning TCP errors and the like instead.

database/postgres.go
```diff
@@ -164,3 +174,77 @@ func PostgresDeadlockFound(err error) bool {
...
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return false // don't retry if the caller's context timed out
+	}
+
+	// pgx wraps connection errors with these messages
+	msg := err.Error()
+	if strings.Contains(msg, "connection reset by peer") ||
+		strings.Contains(msg, "broken pipe") ||
+		strings.Contains(msg, "connection refused") ||
```
elovelan: I'm pretty sure this is the only safe one of this list (and even this we should confirm)

database/postgres.go
```diff
@@ -164,3 +174,77 @@ func PostgresDeadlockFound(err error) bool {
...
+		return true
+	}
+
+	return false
+}
+
+// RetryPostgres executes fn up to maxAttempts times, retrying on transient
+// connection errors. This is intended for use around individual database
+// operations to survive brief outages like AlloyDB maintenance switchovers.
+func RetryPostgres(ctx context.Context, maxAttempts int, fn func() error) error {
```
elovelan: FWIW both [the thread on `pgx` that the author references whenever retries come up](https://github.com/jackc/pgx/issues/701), with [this comment that the author agrees should work](https://github.com/jackc/pgx/issues/701#issuecomment-2446088827) and [PGXTransactions](https://github.com/synaptor/pgxtransactions/blob/main/pgxtransactions.go) recommend using an explicit transaction instead of an implicit one.

Even though both implicit transactions and the commit phase of an explicit _can_ hit the timing issue that @InfernoJJ mentioned in one of our calls, where it successfully commits but the connection throws an error, there are fewer failure scenarios in the latter case. Using an explicit transaction would also allow us to go a step further and capture the `pg_current_xact_id()` before the commit so on failure we could call `pg_xact_status()` to see if the commit was succesful.

Note, neither PgBouncer nor Google's hosted version of it, [Managed Connection Pooling](https://docs.cloud.google.com/alloydb/docs/configure-managed-connection-pooling) solve this problem since a bouncer cannot safely make decisions without a clear commit status either. Google also explicitly calls out in the linked article that a bouncer is only recommended for short-lived tasks, of which we don't have many of, if any.

~~There are also some tuning options that we can do on the AlloyDB side that might be worth adding to make sure a connection is removed from the pool sooner than it is today. I just want to double check some of the suggestions I got before posting them in the ticket.~~ Confirmed these won't actually help

database/postgres.go
elovelan: I [can't find this in the codebase](https://github.com/search?q=repo%3Ajackc%2Fpgx+%22refused%22&type=code)

database/postgres.go
stevemsmith: Pushed a commit to address this issue, includes citations for each error.

database/postgres.go
stevemsmith: Got it.  Pushed a commit to replace the backoff with jitter in a 0–100ms window per retry.

Calling services can control patience through the context timeout. In `connections` for example, the queryScanConnection method can wrap its context with a 1 second deadline before the query runs. If the retry loop hasn't gotten a working connection within that second across all 3 attempts, it stops and returns an error.

The service responds with a 500, same as today, but only after trying to recover from a dead connection rather than giving up on the first failure.

Does this address the concern?  We need to retry as that's dictated by the AlloyDB patching mechanism.

database/postgres.go
stevemsmith: Just saw this - good callout.  Commit pushed along with comment.

database/postgres.go
stevemsmith: I think this issue exists today without any retry logic?  Currently services just get a 500 with no idea whether the write landed.  A risk to be sure, but this change is specifically aimed at retrying dead connections in the pool causing failures.

database/postgres.go
```diff
@@ -164,3 +175,113 @@ func PostgresDeadlockFound(err error) bool {
...
+	}
+
+	// Network-level errors: connection reset, broken pipe, EOF, etc.
+	// These occur when the TCP connection is severed during a switchover.
+	var netErr *net.OpError
+	if errors.As(err, &netErr) {
+		return true
+	}
+	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+		return true
```
elovelan: based on my analysis, there's no way to know if these happened before or after the server accepted the query so I think we should remove these.

database/postgres.go
```diff
@@ -164,3 +175,113 @@ func PostgresDeadlockFound(err error) bool {
...
+	// than preserving the original net.OpError type. These are safe to retry
+	// because they occur on the write path — the server never received the query.
+	// "conn closed" is pgx's own sentinel for a connection already closed before use.
+	// Note: "connection refused" and "unexpected EOF" are intentionally excluded —
+	// the former does not appear in pgx's codebase, the latter is already caught
+	// above by errors.Is(err, io.ErrUnexpectedEOF).
+	msg := err.Error()
+	if strings.Contains(msg, "connection reset by peer") ||
+		strings.Contains(msg, "broken pipe") ||
+		strings.Contains(msg, "conn closed") {
```
elovelan: same issue here as [above](https://github.com/moov-io/base/pull/490/changes#r3482973525)

database/postgres.go
elovelan: > this issue exists today without any retry logic ... with no idea whether the write landed.

Sure, but services could be checking if they get an error calling the database and using one of many methods to retry if it's safe. If not, it's a matter of either:

- explicitly choosing to be alerted instead of designing for idempotency or retry handling
- forgetting the **first** rule of the [fallacies of distributed computing](https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing):

  > 1. The network is reliable

I suspect this library can help with both by having two distinct functions:
   1. one with automatic retries that are guaranteed to be safe.

      By using a transaction, per my suggestion, we can increase the number of situations where this is true, though `fn` should probably take a transaction argument like in [`pgx.BeginTxFunc`](https://pkg.go.dev/github.com/jackc/pgx/v5#BeginTxFunc) (which this function could use under the hood), so it's clear what it's doing.
   2. a similar one with _unsafe_ retries that makes it clear the function being run be idempotent.

stevemsmith:
> Can we try this in some apps before merging?

Yes, I'm running the changes in one service, will try a few others.

InfernoJJ (dismissed):

jwbraucher (approved):

database/postgres.go
```diff
@@ -23,35 +24,47 @@ const (
...
+	poolConfig, err := buildPgxPoolConfig(ctx, config, databaseName)
+	if err != nil {
+		return nil, logger.LogErrorf("building pgx pool config: %w", err).Err()
 	}
 
-	db, err := sql.Open("pgx", connStr)
+	// HealthCheckPeriod makes pgxpool ping idle connections in the background.
+	// Dead connections (e.g. from an AlloyDB switchover) are evicted before
+	// the application ever sees them.
+	poolConfig.HealthCheckPeriod = 1 * time.Second
```
elovelan: this isn't true, all that this checks is [expired and idle](https://github.com/moov-io/base/pull/490/changes#r3506949874) and 1 sec seems like overkill. With a 5 min lifetime, the default of 1 min is probably fine, though 30 sec is also reasonable.

Instead, we should set `config.ShouldPing` to either `return true` (will ping before every call so safest!) or `p.IdleDuration > 300ms` (recommended by Claude, we can confirm)

database/database.go
```diff
@@ -112,3 +112,27 @@ func ApplyConnectionsConfig(db *sql.DB, connections *ConnectionsConfig, logger l
 
 	return db
 }
+
+// ApplyPostgresConnectionsConfig applies connection pool settings with safe defaults
+// for Postgres/AlloyDB. If any value in the provided config is zero, the corresponding
+// default from DefaultPostgresConnectionsConfig is used. This ensures all services get
+// failover-safe pool settings even if they don't explicitly configure them.
+func ApplyPostgresConnectionsConfig(db *sql.DB, connections *ConnectionsConfig, logger log.Logger) *sql.DB {
```
elovelan: the idle settings need to be applied to [the `pgxpool` config](https://pkg.go.dev/github.com/jackc/pgx/v5/pgxpool#Config) during the call to `postgresConnection`. Applying them here will actually cause issues because this is applying against the `sql.DB` and thus [resetting `MaxIdleConns`](https://github.com/stevemsmith/base/blob/fix/alloydb-failover-recovery/database/database.go#L98) from its pgxpool required value of 0. I'm not so sure about `MaxLifetime` either, will have to research.

~~I _think_ it's fine to set `MaxOpen` here.~~

database/postgres.go
```diff
@@ -23,35 +24,47 @@ const (
...
 	if err != nil {
-		_ = db.Close()
+		pool.Close()
 		return nil, logger.LogErrorf("connecting to database: %w", err).Err()
 	}
 
+	// Wrap the pgxpool in a *sql.DB so the rest of the codebase doesn't change.
+	// pgxpool manages the real pool (with health checks); database/sql pool
+	// settings are applied on top via ApplyPostgresConnectionsConfig.
+	db := stdlib.OpenDBFromPool(pool)
```
elovelan: might be worth calling out in a comment that [this sets MaxIdle to 0 on the `sql.DB` when using this, and it's important to not change it](https://pkg.go.dev/github.com/jackc/pgx/v5@v5.10.0/stdlib#GetPoolConnector).

elovelan (commented):
There are a few other tuning things that Claude recommended that I want to understand better, but I wanted to submit this because these comments are the most important to address.

database/database.go
elovelan: setting `MaxLifetime` to `0` will allow `pgxpool` to manage it itself (`config.MaxConnLifetime`), which is actually a little smarter since it adds jitter. My gut says we do the same for `MaxOpen` since `sql.DB` won't try to handle it and will leave it up to `pgxpool`'s `config.MaxConns`

database/postgres.go
elovelan: we might also want to consider setting `config.PingTimeout`, though this is less important since I'm pretty sure a closed connection will fail immediately so this is more for dealing with "hung" connections.

database/database.go
elovelan: Recommendations from Claude and reasoning:

```go
cfg.MinIdleConns = 2           // pre-warm so reconnect latency isn't on the critical path

// Rotation — keep some churn so stale state doesn't accumulate, but you don't
// need aggressive values since AlloyDB's IP is static and reconnects are cheap+fast.
cfg.MaxConnLifetime = 15 * time.Minute   // jitter is applied automatically by pgxpool
cfg.MaxConnIdleTime = 5 * time.Minute
cfg.HealthCheckPeriod = 30 * time.Second // reaper cadence; not a ping, just age/idle eviction

// Liveness — the part that actually matters for rollover.
cfg.PingTimeout = 1 * time.Second        // bound dead-conn detection; alloy disconnects are <1s
cfg.ShouldPing = func(_ context.Context, p pgxpool.ShouldPingParams) bool {
    return p.IdleDuration > 200 * time.Millisecond
}
```

- `PingTimeout = 1s` — this is the single most important knob. A ping against a dead backend usually returns a TCP RST immediately, but a hung/TIME_WAIT peer can hang; cap it so the Acquire retry loop (maxConns+1 attempts) stays inside your request budget. With <1s disconnects, 1s is the right ceiling.
- `ShouldPing = idle > 200ms` — the disconnect can happen at any instant; idle duration doesn't tell you whether a conn crossed it, so the safe choice is to ping nearly always. I gate at `200ms` rather than `return true` to skip the ping on a hot conn that was used moments ago (those are the ones that would die mid-query anyway, which ping can't prevent — only app retry can). At very high QPS where nearly every acquire is `<200ms` idle, set this to `return true` and eat the `SELECT 1` cost; at low QPS, `return true` is fine too. Don't use the default `> 1s` — that's too lax for the `<1s` disconnect guarantee of AlloyDB.
- `MinIdleConns = 2` — keeps a warm floor so the first post-rollover request doesn't pay pgx.ConnectConfig latency (TLS handshake + auth). Cheap insurance.
- `MaxConnLifetime = 15m / MaxConnIdleTime = 5m` — with static AlloyDB IP you don't need short values for DNS rotation. 15m/5m bounds staleness and lets the reaper drop any dead idle conn, without churning. `pgxpool` adds lifetime jitter automatically so you won't get thundering-herd reconnects.
- `HealthCheckPeriod = 30s` — cadence of the background reaper. It's not a ping, so lowering it further just adds eviction frequency, not liveness checks.
- `MaxConns` — the one I can't pick for you. Formula: roughly `2–4 × app CPU cores` per instance, capped by `AlloyDB max_connections ÷ (app instances × safety factor)`. The rollover-relevant constraint is the other direction: don't make it so large that the `Acquire` loop serially burning through N dead idle conns (each ≤ PingTimeout) blows the request budget. At `ShouldPing = idle > 200ms` and `PingTimeout=1s`, `~16–32` is comfortable; hundreds means a worst-case rollover could stack several seconds of ping failures before you get a fresh conn.

jwbraucher (commented):
reverting my approval - I agree with @elovelan that the healthcheck does not include a ping()

stevemsmith:
Removed `DefaultPostgresConnectionsConfig` and `ApplyPostgresConnectionsConfig` to keep it simpler.

Connection lifetime and pool size tuning are out of scope for this PR, the goal is dead connection detection, not pool rotation policy.  These can be revisited in a follow-up if needed.

jwbraucher (approved):

elovelan (approved):
looks great! 🥳

adamdecaf:
/gemini review

@gemini-code-assist please re-review the current PR head (ShouldPing + pgxpool path, retry logic removed). Focus on connection lifecycle (pool/dialer close), config mapping, and failover behavior.

database/postgres.go
elovelan: @stevemsmith, why did you mark this resolved as well as [my related comments](https://github.com/moov-io/base/pull/490/changes#r3507010696)?

database/postgres.go
```diff
@@ -20,38 +23,167 @@ const (
 	// https://www.postgresql.org/docs/current/errcodes-appendix.html
 	postgresErrUniqueViolation = "23505"
 	postgresErrDeadlockFound   = "40P01"
+
+	// Bound ShouldPing wait so acquire retries during failover stay inside
+	// typical request budgets. AlloyDB disconnects usually fail fast (TCP RST);
+	// this caps hung/TIME_WAIT peers.
+	defaultPostgresPingTimeout = time.Second
```
atonks2, line 31: `1s` here might be a little high. 

`Pool.Acquire` retries up to `MaxConns + 1` times (26 with this new default), and each retry can burn the full ping timeout. So when the connection is blackholed (dropped, not rejected), it means one acquire can sit there for ~26s, bounded only by the caller's context. 

Maybe it would be good to drop this to 200–300ms? The comment says AlloyDB usually RSTs, which is true, but this timeout only matters in the case where it doesn't.

adamdecaf (dismissed):

InfernoJJ (dismissed):

database/database.go
```diff
@@ -87,6 +91,19 @@ func DeadlockFound(err error) bool {
...
 
+// ApplyPostgresConnectionsConfig applies connection pool settings onto a *sql.DB.
+//
+// Deprecated: Postgres connections from New use pgxpool under the hood. Pool
+// settings are applied via ApplyPostgresPoolConfig inside postgresConnection.
+// Calling this on a Postgres *sql.DB from New is incorrect: SetMaxIdleConns
+// with a non-zero value breaks OpenDBFromPool, and the other setters do not
+// configure the underlying pgxpool. Prefer ConnectionsConfig on PostgresConfig
+// (applied automatically) or ApplyPostgresPoolConfig when building a pool.
+func ApplyPostgresConnectionsConfig(db *sql.DB, connections *ConnectionsConfig, logger log.Logger) *sql.DB {
```
atonks2, line 102: If this is deprecated, should we include it in this PR at all?

adamdecaf:
Pushed a follow-up onto this branch and refreshed the PR description to match head:

1. **Description** — documents ShouldPing, pgxpool mapping, defaults (MaxOpen=25 / MaxLifetime=5m / MaxIdleTime=2m), deprecated helper, no retry, and operator upgrade notes.
2. **Operator docs** — `PostgresConfig`, `ConnectionsConfig`, and `DefaultPostgresConnectionsConfig` godoc call out the behavior change vs unlimited `database/sql` zeros.
3. **Unit tests** — `database/postgres_pool_test.go` covers `ResolvePostgresConnectionsConfig`, `ApplyPostgresPoolConfig`, and `openDBFromPool` Close lifecycle (pool closed, idempotent Close).

```text
go test ./database/ -run 'TestResolvePostgresConnectionsConfig|TestApplyPostgresPoolConfig|TestOpenDBFromPool|TestConnectionsConfigOrder' -count=1
```

adamdecaf (dismissed):

adamdecaf (approved):
CI is green after gosec G115 fix (clamp MaxOpen before int32 conversion).

database/model_config.go
```diff
@@ -104,3 +125,35 @@ type RetryConfig struct {
...
+// # Operator notes
+//
+// These defaults change runtime behavior versus the previous database/sql +
+// pgx driver path, where unset fields meant unlimited:
+//
+//   - MaxOpen (25): pgxpool always has a finite MaxConns. Unset MaxOpen no
+//     longer means unlimited — it becomes 25. High-concurrency services must
+//     set PostgresConfig.Connections.MaxOpen explicitly if they need more.
+//   - MaxLifetime (5m): connections are recycled after this age (pgxpool
+//     adds jitter). Unset previously meant no max lifetime. Shorter lifetime
```
atonks2, line 141: Are you sure about this? My understanding is that pgxpool defaults jitter to 0.

atonks2:
Also, I think we're going to lose db.Stats() once this rolls out. It won't return the stats from the underlying connection pool.

database/postgres.go
adamdecaf: Oh yea, we should lower that.

adamdecaf: @atonks2 (replying as **Grok**, not Adam — I reviewed this after merge)

You're right about the worst-case math. `pgxpool.Pool.Acquire` retries up to `maxConns + 1` times, and each failed `ShouldPing` can burn the full `PingTimeout`:

https://github.com/jackc/pgx/blob/v5.10.0/pgxpool/pool.go#L610-L644

With the new default `MaxOpen=25` and `PingTimeout=1s`, a blackholed peer path is theoretically ~26s of serial pings before giving up (or earlier if the caller's context cancels). AlloyDB's usual TCP RST still fails fast, so the happy/failover path is fine — but the timeout only matters when it *doesn't* RST, which is exactly when 1s hurts.

I'd support dropping the default to something like **250–300ms** in a follow-up. Still enough headroom for a real ping RTT, and it keeps a full-pool dead-conn burn inside a more typical request budget even without relying on context cancellation.

database/database.go
adamdecaf: @atonks2 (Grok here, not Adam)

Fair question. Looking at it cold: keeping a **deprecated** helper that still calls `ApplyConnectionsConfig` (including `SetMaxIdleConns`) is more footgun than compatibility shim. On the Postgres path from `New`, any non-zero idle setting on `*sql.DB` breaks the pgxpool handoff (`MaxIdleConns` must stay 0).

I'd lean toward **removing it** in a cleanup follow-up unless something outside this repo still imports it. Deprecation without a grace period of real callers doesn't buy much, and the correct APIs are already `PostgresConfig.Connections` / `ApplyPostgresPoolConfig`.

database/model_config.go
adamdecaf: @atonks2 (Grok, not Adam)

You're correct — that godoc line is wrong as written.

In pgx v5.10, `MaxConnLifetimeJitter` defaults to **0** (`pool_max_conn_lifetime_jitter` default 0). Jitter is only applied when configured:

```go
jitterSecs := rand.Float64() * config.MaxConnLifetimeJitter.Seconds()
maxAgeTime := time.Now().Add(config.MaxConnLifetime).Add(...)
```

So “pgxpool adds jitter” is not true out of the box. Worth fixing the comment on master, and optionally setting an explicit jitter (e.g. ~10% of `MaxConnLifetime`, or a fixed 30s) in `ApplyPostgresPoolConfig` so lifetime recycling doesn’t thundering-herd after deploy/rollover.

adamdecaf:
@atonks2 — following up on your `db.Stats()` note (replying as **Grok**, not Adam).

You're right: after this change Postgres is `pgxpool` under `OpenDBFromPool` with `MaxIdleConns=0` on the `*sql.DB` wrapper. Connections are not retained in `database/sql`, so `db.Stats()` only sees the empty handoff layer (open/idle/inuse stay ~0). Real pressure lives on `pool.Stat()` (`TotalConns` / `AcquiredConns` / `IdleConns` / `EmptyAcquireCount` / lifetime+idle destroy counts, etc.).

That means `sql.ObserveDB` / `MeasureStats` would have gone dark for Postgres capacity metrics if left as-is.

Opened a follow-up that registers a `StatsProvider` from `pgxpool.Stat` so the existing `sql_connections*` series keep working, plus `database.PoolStat` for direct access:

https://github.com/moov-io/base/pull/505

adamdecaf:
@atonks2 (Grok, not Adam) — follow-up on the `db.Stats()` thread:

You were right that go-libs is the real consumer: [`observability/sql.MeasureStats`](https://github.com/moovfinancial/go-libs/blob/main/observability/sql/sql.go) already puts pool stats on a periodic OTel **`db-metrics`** span via `db.Stats()`.

Rather than scraping Prometheus inside base, we:

1. **base** exposes `database.PoolDBStats` from the underlying pgxpool → https://github.com/moov-io/base/pull/507  
2. **go-libs** prefers that in `MeasureStats` → https://github.com/moovfinancial/go-libs/pull/92  

(The earlier Prometheus-only #505 was closed in favor of this.)

PR #509, opened 2026-08-08
fix(http): require CORS origin allowlist for credentialed responses

SashaMIT:
## Summary
- `SetAccessControlAllowHeaders` previously reflected any `https://` Origin with `Access-Control-Allow-Credentials: true` (and allowed Cookie / X-User-Id).
- Require an explicit allowlist via `MOOV_CORS_ALLOW_ORIGINS` (comma-separated exact Origins).
- Keep `http://localhost:<port>` allowed for local development.
- Update tests; add coverage that unlisted https Origins get no ACAO / credentials headers.

## Test plan
- [x] `go test ./http/ -count=1`
- [x] Allowlisted Origin still receives CORS headers
- [x] Unlisted `https://evil.example` receives none
- [x] Localhost still allowed

## Deploy note
Set `MOOV_CORS_ALLOW_ORIGINS=https://moov.io,...` (exact browser Origins) in environments that rely on credentialed browser calls.

### DIFF
--- a/http/response_test.go
+++ b/http/response_test.go
@@ -21,8 +21,12 @@ var (
 )
 
 func TestResponse__Wrap(t *testing.T) {
+	t.Setenv(CORSAllowedOriginsEnv, "https://moov.io")
+	ResetCORSAllowlistForTest()
+	t.Cleanup(ResetCORSAllowlistForTest)
+
 	req := httptest.NewRequest("GET", "https://api.moov.io/v1/ach/ping", nil)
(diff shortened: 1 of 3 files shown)


adamdecaf:
Landed on master as 980aa65 (amended for reusable SetCORSAllowedOrigins / OriginAllowedForCORS API). Services should bump moov-io/base and set MOOV_CORS_ALLOW_ORIGINS — no local CORS forks needed.

PR #512, opened 2026-08-12
fix(randx): use upper bound in Between

SashaMIT:
## Problem

`randx.Between(lower, upper)` never reads `upper`. It passes `lower` as the exclusive maximum to `crypto/rand.Int` and then adds `lower` to the result, so the returned value falls in `[lower, 2*lower)` instead of `[lower, upper)`.

The existing test only passes because `[100, 200)` loosely overlaps the asserted range of `[100, 250]`.

Additionally, when `lower == 0`, `crypto/rand.Int` is called with a maximum of 0 and panics, so any caller doing `Between(0, n)` crashes the process.

## Fix

- Compute the range as `upper - lower`, draw from `[0, upper-lower)`, and shift the result by `lower`, yielding a uniform value in `[lower, upper)`.
- Return an explicit error when `upper <= lower` instead of panicking inside `crypto/rand`. This matches the package's existing style where `Between` returns errors and `Must` panics on them.

## Tests

- `TestBetween` and `TestMust` now assert both bounds tightly over 1000 samples (`lower <= n < upper`).
- `TestBetween_ZeroLower` covers `lower == 0`, which previously panicked.
- `TestBetween_InvalidBounds` and `TestMust_Panics` cover the `upper <= lower` error path.

Verified that the new tests fail against the old implementation (the zero-lower test panics in `crypto/rand.Int`).

### DIFF
--- a/randx/randx.go
+++ b/randx/randx.go
@@ -2,12 +2,18 @@ package randx
 
 import (
 	"crypto/rand"
+	"fmt"
 	"math/big"
 )
 
-// Between will return a randomly generated integer within the lower and upper bounds provided.
+// Between will return a randomly generated integer within the lower (inclusive)
+// and upper (exclusive) bounds provided.
 func Between(lower, upper int) (int64, error) {
-	n, err := rand.Int(rand.Reader, big.NewInt(int64(lower)))
+	if upper <= lower {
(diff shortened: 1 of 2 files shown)


adamdecaf (approved):

SashaMIT:
Thanks for the merge.

PR #514, opened 2026-08-16
fix(http): set Problem Content-Type before WriteHeader

SashaMIT:
## Summary
`Problem` documented that it sets `Content-Type: application/json`, but it called `WriteHeader` first. `net/http` ignores `Header().Set` after that, so the JSON type never reached the client.

`httptest.ResponseRecorder` does not freeze headers on `WriteHeader`, which is why `TestHTTP__Problem` stayed green. The new test uses a writer that matches the real contract.

## Test plan
- [x] `go test ./http/` green
- [x] `TestHTTP__ProblemSetsContentTypeBeforeWriteHeader` fails if the `Set` is moved back after `WriteHeader`

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -32,8 +32,8 @@ func Problem(w http.ResponseWriter, err error) {
 	if err == nil {
 		return
 	}
-	w.WriteHeader(http.StatusBadRequest)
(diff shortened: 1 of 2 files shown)


adamdecaf (approved):
Looks good. Simplified the test to assert the Content-Type a real client receives.

